PR #22, opened 2016-12-14
use non-deprecated KeyboardEvent methods

### LINKED ISSUE #20: on-key triggers on shift
In Racket, does `on-key` trigger when you press shift? I thought it didnt (will test later). Anyhow, RacketScript does and it makes it hard to enter capital letters.

stchang:
- fixes #20
- fixes Firefox hyphen issue

There may still be some issues but I tested it on all our example games.

### DIFF
--- a/racketscript-extras/racketscript/htdp/universe.rkt
+++ b/racketscript-extras/racketscript/htdp/universe.rkt
@@ -293,51 +293,63 @@
 
 
 ;; TODO: A JS object would be faster.
+;; maps JS KeyboardEvent.key to big-bang KeyEvent
 (define *key-table*
-  (hasheq 16 "shift"
-          17 "control"
-          19 "pause"
-          27 "escape"
-          33 "prior"
-          34 "next"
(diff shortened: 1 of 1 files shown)


stchang:
I used the information here to determine deprecation: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent

vishesh:
Awesome! Regarding [Right/Left]Ctrl/Shift, the docs here[1] lists that there is ShiftLeft/ShiftRight/.... Anything I'm missing?

[1] https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code#Code_values

stchang:
Yeah, I tried using `KeyboardEvent.code` first, but it was producing something like "KeyM" for basic keys, so I used `.key`, which required less translation. I guess I could special case the shift keys. I'll update the pr.

stchang:
ok I handled left/right shifts and squashed the commits

PR #37, opened 2016-12-20
remove unneeded namespace-syntax-introduce

stchang:
Not 100% sure but `expand` [seems to already apply `namespace-syntax-introduce`](http://docs.racket-lang.org/reference/Expanding_Top-Level_Forms.html?q=expand#%28def._%28%28quote._~23~25kernel%29._expand%29%29)?

All the tests still pass if we take it out, and other example programs seem to work as well.


vishesh:
Sounds right! Merged.

PR #43, opened 2017-01-05
add remaining (currently supported) list tests

stchang:
- to support rest of racket/list, we need:
  - with-handlers
  - optional and kw args
  - procedure-arity-includes and friends

Other edits:
- fix makeInit to return Vector instead of Array
- Numbers.add missing initial 0
- make-sequence needs extra arg for Racket >= 6.7.0.4

### DIFF
--- a/racketscript-compiler/racketscript/compiler/runtime/kernel.js
+++ b/racketscript-compiler/racketscript/compiler/runtime/kernel.js
@@ -64,6 +64,10 @@ exports["number?"] = Core.Number.check;
 
 exports["integer?"] = Number.isInteger;
 
+exports["real?"] = function (v) {
+    return Core.Number.check(v);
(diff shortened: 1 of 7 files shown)


racketscript-compiler/racketscript/compiler/runtime/kernel.js
```diff
@@ -64,6 +64,11 @@ exports["number?"] = Core.Number.check;
 
 exports["integer?"] = Number.isInteger;
 
+// TODO: can js have non-real nums?
+exports["real?"] = function (v) {
+    return true;
```
vishesh: Core.Number.check(v)

All JS numbers are double precision floats.

stchang:
Thanks. I fixed `real?`

Issue #44, opened 2017-01-06
Migrate kernel.js to RacketScript


vishesh:
Try to write as much runtime in RacketScript as possible.

https://github.com/vishesh/racketscript/compare/task/rjs-kernel
https://github.com/vishesh/racketscript/tree/task/rjs-kernel

- [X] Add a minimal `#lang racketscript/boot` for writing kernel.
- [X] Migrate basic primitives to kernel.rkt
- [X] Handle cyclic dependency for compiling kernel
- [ ] Add checks by implementing a minimal contract system
- [ ] Migrate all kernel.js to RacketScript
- [ ] Optimization to use native operator doesn't work in kernel.rkt

stchang:
Nice. Is it worth it to keep adding tests? Or should I wait until the kernel is ported?

vishesh:
If you have made some changes since then, we can mege that. Otherwise, its probably good idea to wait for couple more days for this to shape up and avoid some JavaScript :)

PR #46, opened 2017-01-07
Migrate kernel.js/paramz.js/unsafe.js to RacketScript

vishesh:
- Implement a minimal #lang racketscript/boot for writing kernel.js/...
-  For these primitive modules, avoid possible cycles automatically. This happens because certain macro introduced bindings would refer to original `#%kernel`
- Additions to interop and IL for handling `undefined`/`null` 

ISSUE #44


vishesh:
@stchang All tests pass. This is a big patch, but if you get some time, please take a quick look. Mainly `racketscript-compiler/racketscript/boot.rkt` and `kernel.rkt`.

I've ported most JS code in kernel to RacketScript. What's left is Regex and few stuff you added recently. We just point to kernel.js implementation in that case.

Tests will run 2x slower for now :(

racketscript-compiler/racketscript/compiler/runtime/kernel.rkt
```diff
@@ -0,0 +1,824 @@
...
+(define+provide (struct-type? v)
+  (#js.Core.Struct.isStructType v))
+
+;; --------------------------------------------------------------------------
+;; Vectors
+
+(define+provide vector
+  (v-λ ()
+     (#js.Core.Vector.make (#js.Core.argumentsToArray arguments) #t)))
+
```
stchang: add comment indicating optional arg

racketscript-compiler/racketscript/compiler/runtime/kernel.rkt
```diff
@@ -0,0 +1,824 @@
...
+(define+provide (string-length v)
+  #js.v.length)
+
+(define+provide (string-downcase v)
+  (#js.v.toLowerCase v))
+
+(define+provide (string-upcase v)
+  (#js.v.toUpperCase v))
+
+(define+provide (substring str start end)
```
stchang: add comment for optional arg

avelino:
PR merged, I think you can close this issue

PR #73, opened 2017-08-25
Various fixes and improvements

glebm:
1. Fixes the assembler to compare to `!== false` in `if` statements.
   This fixes the behaviour of various functions such as `and`, `if`,
   and `cond` to match Racket, where `#f` is the only "falsy" value.
   **Update:** This change has been removed from this PR, see https://github.com/vishesh/racketscript/pull/73#discussion_r135356677

2. Adds support for `~x` to `format`.
3. Fixes `error` argument formatting.
4. Fixes the default contract check message.
5. Allows passing an `out` argument to `display`, `newline`, and `print`.
6. Adds basic implementations of the following functions:

   * `string-ref`
   * `make-string`
   * `list->string`
   * `char?` and `char` comparison functions
   * `char->integer`
   * `open-output-string`, `get-output-string`, and `string-port?`

This actually makes Waxeye work (the Unicode issue remains, but the other stuff seems to work).
Waxeye demo: https://glebm.github.io/waxeye/demo.html

### DIFF
--- a/racketscript-compiler/racketscript/compiler/runtime/core/ports.js
+++ b/racketscript-compiler/racketscript/compiler/runtime/core/ports.js
@@ -1,73 +1,99 @@
 import {Primitive} from "./primitive.js";
 import * as $ from "./lib.js";
 
-//TODO: Quick hacked up implementation around console.log
-class Port extends Primitive {
-    constructor(read, write) {
-	super();
-	this.__state = {
-	    buffer: []
-	} ;
-	this.__read = read;
-	this.__write = write;
-    }
-
-    read(n_char) {
-	if (this.__read === undefined) {
-	    throw $.racketCoreError("Not Implemented");
-	} else {
-	    return this.__read(this.__state, n_char);
-	}
-    }
-
-    write(chars) {
-	if (this.__write === undefined) {
-	    throw $.racketCoreError("Not Implemented");
-	} else {
-	    return this.__write(this.__state, chars);
-	}
-    }
-
+class AbstractPort extends Primitive {
     isOutputPort() {
-	return this.__write && true;
+        return false;
     }
 
     isInputPort() {
-	return this.__read && true;
+        return false;
     }
 }
 
-export function makeOutputPort(write) {
-    return new Port(undefined, write);
-}
-
-export function makeInputPort(read) {
-    return new Port(read);
-}
-
-export function check(v) {
-    return (v instanceof Port);
-}
-
-export function checkInputPort(v) {
-    return check(v) && v.isInputPort();
-}
-
-export function checkOutputPort(v) {
-    return check(v) && v.isOutputPort();
-}
-
-export let standardOutputPort = makeOutputPort((state, chars) => {
-    let nl_index = chars.lastIndexOf("\n");
-    if (nl_index >= 0) {
-	let flushchars = state.buffer.join("") + chars.slice(0, nl_index);
-	let rest_chars = chars.slice(nl_index + 1);
-	state.buffer = [];
-	if (rest_chars !== "") {
-	    state.buffer.push(rest_chars);
-	}
-	console.log(flushchars);
-    } else {
-	state.buffer.push(chars);
+class AbstractOutputPort extends AbstractPort {
+    write() {
+        throw new Error(`Expected ${this.constructor.name} to implement write(chars)`)
     }
-});
+
+    isOutputPort() {
+        return true;
+    }
+}
+
+export function isPort(v) {
+    // When compiling with traceur, the `instanceof` check would fail,
+    // due to a limitation of traceur.
+    // TODO: Replace with `v instanceof AbstractPort` once this is fixed.
+    return v && (
+        v.constructor === NewlineFlushingOutputPort ||
+        v.constructor === OutputStringPort);
+}
+
+export function isInputPort(v) {
+    return isPort(v) && v.isInputPort();
+}
+
+export function isOutputPort(v) {
+    return isPort(v) && v.isOutputPort();
+}
+
+// Only writes output via the given `writeFn` when encountering a newline,
+// othewise buffers the output.
+class NewlineFlushingOutputPort extends AbstractOutputPort {
+    constructor(writeFn) {
+        super();
+        this._buffer = [];
+        this._writeFn = writeFn;
+    }
+
+    write(chars) {
+        const lastNewlineIndex = chars.lastIndexOf("\n");
+        if (lastNewlineIndex >= 0) {
+            let flushchars = this._buffer.join("") + chars.slice(0, lastNewlineIndex);
+            let restChars = chars.slice(lastNewlineIndex + 1);
+            this._buffer = [];
+            if (restChars !== "") {
+                this._buffer.push(restChars);
+            }
+            this._writeFn(flushchars);
+        } else {
+            this._buffer.push(chars);
+        }
+    }
+}
+
+export const standardOutputPort = new NewlineFlushingOutputPort((str) => console.log(str));
+export let currentOutputPort = standardOutputPort;
+
+export const standardErrorPort = new NewlineFlushingOutputPort((str) => console.log(str));
+export let currentErrorPort = standardErrorPort;
+
+
+class OutputStringPort extends AbstractOutputPort {
+    constructor() {
+        super();
+        this.__buffer = [];
+        this.__cachedString = null;
+    }
+
+    write(chars) {
+        this.__buffer.push(chars);
+        this.__cachedString = null;
+    }
+
+    getOutputString() {
+        if (this.__buffer.length > 1) {
+            this.__buffer = [this.__buffer.join('')];
+        }
+        return this.__buffer[0];
+    }
+}
+
+export function openOutputString() {
+    return new OutputStringPort();
+}
+
+export function getOutputString(outputStringPort) {
+    return outputStringPort.getOutputString();
+}
--- a/racketscript-compiler/racketscript/compiler/runtime/kernel.js
+++ b/racketscript-compiler/racketscript/compiler/runtime/kernel.js
@@ -1,15 +1,25 @@
 import * as Core from "./core.js";
 
+/* --------------------------------------------------------------------------*/
+// String construction and manipulation
+
 export function format(pattern, ...args) {
-    //TODO: Only ~a is supported
+    //TODO: Only ~a and ~x are supported
     var matched = 0;
-    return pattern.replace(/~a/g, function(match) {
-	if (args[matched] == 'undefined') {
-	    throw Core.racketContractError("insufficient pattern arguments");
-        } else {
-            return args[matched++];
+    return pattern.replace(/~[ax]/g, function (match) {
+        if (matched >= args.length) {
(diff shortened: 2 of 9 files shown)


racketscript-compiler/racketscript/compiler/runtime/core/pair.js
```diff
@@ -98,6 +98,10 @@ export function listToArray(lst) {
     return r;
 }
 
+export function listToString(lst) {
```
vishesh: `list->string` takes argument which obeys `(listof char?)` contract. Since the function is not generic enough, I think it should goto `kernel.rkt` where you are implementing the `list->string` function. You probably can just inline the definition there, unless you think this function can be useful otherwise.

racketscript-compiler/racketscript/compiler/runtime/core/ports.js
```diff
@@ -57,6 +57,33 @@ export function checkOutputPort(v) {
     return check(v) && v.isOutputPort();
 }
 
+class OutputStringPort {
```
vishesh: Can you re-write this like we have implemented `standardOutputPort` here[1]? The `Port` class constructor takes a reader function and writer function. Here are the contracts:

```
read: (State, Int) -> Listof<X> [X is Char, Bytes whatever ...]
```
State is the internal mutable state that you can use to keep context and the second argument is the number of characters you want to read.

```
write: (State, Listof<X>) -> Void
```
State is same as for read. The second argument is just list of char, bytes you want to write.

The standard output implemented here[1] uses `console.log` to this (not great but for sake of printing to browser console). Similar approach can be used for making `Ports` to write to DOM or files etc.

Let me know if you find some better approach to address this whole port problem. This file is fairly small, and doesn't have much code to change so I'm open to new ideas.

[1] https://github.com/glebm/racketscript/blob/78d1d34f96bd0c1dedc0a67c775814b29bebd159/racketscript-compiler/racketscript/compiler/runtime/core/ports.js#L87

racketscript-compiler/racketscript/compiler/runtime/kernel.rkt
```diff
@@ -613,6 +616,10 @@
 (define+provide format #js.Kernel.format)
 (define+provide symbol? #js.Core.Symbol.check)
 
+(define+provide (make-string n c)
```
vishesh: Please add a `TODO` note that in future this needs to return a mutable string. For now its OK, as we don't have that primitive anyway.

racketscript-compiler/racketscript/compiler/runtime/kernel.rkt
```diff
@@ -657,6 +662,27 @@
 (define+provide (string-split str sep)
   (#js.Core.Pair.listFromArray (#js.str.split sep)))
 
+(define+provide (char? c)
```
vishesh: Please add a marker like we do for other primitives for organization.
```
;; --------------------------------------------------------------------------
;; Characters
```

racketscript-compiler/racketscript/compiler/runtime/kernel.js
```diff
@@ -33,7 +35,6 @@ export function display(v, out) {
 
 export function print(v, out) {
     /* TODO: this is still line */
-    out = out || Core.Ports.standardOutputPort;
```
vishesh: Can you confirm if print is not broken when out is not given? It should use standard output port if nothing is given.

racketscript-compiler/racketscript/compiler/runtime/kernel.js
```diff
@@ -1,23 +1,25 @@
 import * as Core from "./core.js";
 
 export function format(pattern, ...args) {
-    //TODO: Only ~a is supported
+    //TODO: Only ~a and ~x are supported
```
vishesh: Please add tests for this if you got some time. Look here[1] to get started. Let me know if the documentation is insufficient to be able to figure out how to run individual test. The test runner program is `fixture.rkt` in tests directory.

[1] https://github.com/glebm/racketscript/tree/78d1d34f96bd0c1dedc0a67c775814b29bebd159/tests

racketscript-compiler/racketscript/compiler/runtime/kernel.rkt
```diff
@@ -657,6 +662,27 @@
...
+  (typeof #js.c "string"))
```
vishesh: Also can you please add some tests for the character functions you added?

racketscript-compiler/racketscript/compiler/assembler.rkt
```diff
@@ -156,7 +156,7 @@
     [(ILIf expr t-branch f-branch)
      (emit "if (")
      (assemble-expr expr out)
-     (emit ") {")
+     (emit " !== false) {")
```
vishesh: We should not modify assembler for fixing if/else. Fundamentally, the IL language obeys most of JavaScript semantics. So in `transform.rkt`[1] you should generate IL equivalent of this.

It would be nicer to have this change in separate patch.

[1] https://github.com/glebm/racketscript/blob/78d1d34f96bd0c1dedc0a67c775814b29bebd159/racketscript-compiler/racketscript/compiler/transform.rkt#L293

glebm: Could you perhaps do this change in a separate commit and I'll remove this change from here?
I don't understand this part well enough to fix it in the right place.

By the way, in the snippet you've linked:

```
    [(If pred-e t-branch f-branch)
     (define-values (ps pe) (absyn-expr->il pred-e #f))
     (define-values (ts te) (absyn-expr->il t-branch overwrite-mark-frame?))
     (define-values (fs fe) (absyn-expr->il f-branch overwrite-mark-frame?))
```

What do these stand for: `il`, `absyn`,  `s`/`e`?

vishesh: Sure I can do the change. I think you also asked what `absyn` and `il` stands for. Both are simple abstract language which pretty much corresponds to grammar or Racket and JavaScript respectively. Absyn[1] is pretty much like Racket and IL[2] (Intermediate Language) is pretty much like JavaScript. Since IL is more like JavaScript, I asked to make that change in Absyn which generate IL. Another reason we want to keep IL more like JavaScript is to provide a more flexible FFI, that lets you produce almost raw JS without ditching Racket.

[1] https://github.com/vishesh/racketscript/blob/master/racketscript-compiler/racketscript/compiler/absyn.rkt
[2] https://github.com/vishesh/racketscript/blob/master/racketscript-compiler/racketscript/compiler/il.rkt

vishesh:
BTW great work on Waxeye! I saw the demo and I'm really glad to see something cool working on top of RacketScript.

racketscript-compiler/racketscript/compiler/runtime/kernel.rkt
glebm: Done

racketscript-compiler/racketscript/compiler/runtime/core/pair.js
glebm: Done

racketscript-compiler/racketscript/compiler/runtime/kernel.rkt
glebm: Done

racketscript-compiler/racketscript/compiler/runtime/kernel.js
glebm: Done

racketscript-compiler/racketscript/compiler/runtime/kernel.rkt
glebm: Done

racketscript-compiler/racketscript/compiler/runtime/kernel.rkt
glebm: Done

racketscript-compiler/racketscript/compiler/runtime/kernel.rkt
glebm: Done

racketscript-compiler/racketscript/compiler/runtime/kernel.js
glebm: Confirmed, it does use the `current-output-port` (looks like default arguments in `define+provide` work).

racketscript-compiler/racketscript/compiler/runtime/core/ports.js
glebm: I've refactored the entire file to use classes + inheritance, which seem to be the right tool in this case.

racketscript-compiler/racketscript/compiler/runtime/core/pair.js
glebm: Done

racketscript-compiler/racketscript/compiler/assembler.rkt
glebm: Thanks, I've removed the changes from this PR, look forward to looking at your fix.

racketscript-compiler/racketscript/compiler/runtime/kernel.js
vishesh: Thanks.

racketscript-compiler/racketscript/compiler/runtime/core/ports.js
```diff
@@ -1,73 +1,100 @@
...
-//TODO: Quick hacked up implementation around console.log
-class Port extends Primitive {
-    constructor(read, write) {
-	super();
-	this.__state = {
-	    buffer: []
-	} ;
-	this.__read = read;
-	this.__write = write;
+class AbstractPort extends Primitive {
```
vishesh: Thanks for the changes here! Here is some feedback:

1. Please fix indentation of this file (4 spaces). I think right now it is 1 tab (or 8 spaces).
2. Parameterized `NewlineFlushing*Port` is definitely a cool change!
3. I think one of the reasons I used passing functions instead of inheritance, is so that we don't have to repeat implementation of or `InputPort*` and `OutputPort* ports in` `InputOutputPort* ` ports. Since ES6 doesn't support mixins.  We can either use embedding pattern to parameterize  the read/write of ports or if you have some ideas.

tests/basic/char.rkt
```diff
@@ -1,4 +1,37 @@
...
+#lang racket/base
 
 (define v #\u5c)
 (displayln v)
+
+(displayln (char? #\a))
+(displayln (char? 1))
+(displayln (char? '(#\a)))
+
+; TODO: This test fails because we do not yet have a separate type for Char.
```
vishesh: Thanks for adding negative cases!

racketscript-compiler/racketscript/compiler/runtime/core/ports.js
vishesh: Agreed, composition would be easier to follow. I think it would also make the predicates for checking input/output/generic ports back to single type check.

glebm: Identation fixed!
Didn't realize vscode does tabs/spaces auto-detection based on what's already in the file by default.

glebm:
Commits squashed, good to merge

PR #86, opened 2017-08-29
dont run optimize tests when optimizations are off

stchang:
The `tests/optimize/tail-sum.rkt` integration test is failing for me when optimizations are off.

Should the tests in `tests/optimize` be run when there are no optimizations enabled?

ps - Loving the activity in this repo.

### DIFF
--- a/tests/fixture.rkt
+++ b/tests/fixture.rkt
@@ -241,7 +241,9 @@
       (displayln "---------------------------------")
       (displayln "::: Optimizations on ::: none :::")
       (displayln "---------------------------------")
-      (run-tests tc-search-patterns)))
+      (run-tests (filter-not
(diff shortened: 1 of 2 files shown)


vishesh:
Thanks! I was thinking why this didn't fail earlier, and seems like I had [this line commented out](https://github.com/vishesh/racketscript/blob/master/tests/optimize/tail-sum.rkt#L11). Can you also uncomment that line with this patch?

stchang:
Ok done. The 3rd case was also failing for me but I have some old versions of things so maybe my default stack size is smaller?

vishesh:
Merged. Good to see you back!

vishesh:
Yeah probably the stack was smaller. You were using an older Debian IIRC? These days I think by default it is around 10k.

PR #87, opened 2017-08-29
delete temporary racket-list test file

stchang:
The `tests/racket-core/racket-list.rkt` file was temporary bc racketscript could not find the racket collects at the time, but this seems to be working now.


stchang:
Oops, not sure why this is passing my machine. Let me investigate.

stchang:
Oh, it looks like `index-of` was added in v6.7. I'll see if I can add a version check to the tests.

vishesh:
Oh. If its too much trouble we can forget about 6.4? Is it wrong to assume that most Racket developers will probably use a version less than a year old?

But yes, it would be good to have a way to check versions as in future we may just get something like this between single release.

stchang:
I added a quick hack that seems to do the trick but I see you already have a "skipped tests" framework. Let me see if I can adapt that to support versioning.

PR #88, opened 2017-08-29
Re-implement hashing for better performance

glebm:
This implementation is cleaner, better documented, and considerably faster than the previous one, as per the microbenchmarks linked in the comments. For example, `hashNumber` is **175x faster**.

The previous code was slow, in part as a result of supporting browsers that are now obsolete, and had branches that were no longer hit after the hashing overhaul in the Unicode PR.

The `smi` function has been replaced with `~~h` because on 64-bit platforms it's not necessary, as the SMI size is 32 bits:
https://github.com/v8/v8/blob/fdfb8c9efbbd96cde5e4cc09cd22146d75e5598b/include/v8.h#L8773

Since all the Facebook (ImmutableJS) code has been removed, this repo is no longer encumbered by the Facebook's PATENTS clause, which made using it impossible or difficult in a lot of companies.


racketscript-compiler/racketscript/compiler/runtime/core/algo/hashing.js
```diff
@@ -0,0 +1,140 @@
...
+ */
+export function hashString(s) {
+    let h = 0;
+    const n = s.length;
+    for (let i = 0; i < n; ++i) {
+        // Benchmarks of various ways to do this:
+        // https://run.perf.zone/view/String-Hashing-Performance-1504040177726
+        h = ~~(((h << 5) - h) + s.charCodeAt(i));
+    }
+    return h;
```
vishesh: Can you point to the source where you got this hash function. I don't recall ever coming across this.

racketscript-compiler/racketscript/compiler/runtime/core/algo/hashing.js
```diff
@@ -0,0 +1,140 @@
+/**
```
vishesh: Content of this file should goto `core/hash_code.js`.

racketscript-compiler/racketscript/compiler/runtime/core/algo/hashing.js
```diff
@@ -0,0 +1,140 @@
...
+    // For small numbers, return the number directly.
+    // This slightly increases the potential number of collisions
+    // with large numbers and floats, but increases the performance by 20%.
+    if (~~n === n) {
+        // If `n` is -0, the above check will pass.
+        // `~~` is here only to convert the potential -0 to 0.
+        return ~~n;
+    }
+    kBufAsF64[0] = n;
+    return kBufAsI32[0] ^ kBufAsI32[1];
```
vishesh: Also this.

racketscript-compiler/racketscript/compiler/runtime/third-party/hash.js
```diff
@@ -1,209 +0,0 @@
...
-	    stringHashCache = {};
-	}
-	STRING_HASH_CACHE_SIZE++;
-	stringHashCache[string] = hash;
-    }
-    return hash;
-}
-
-// http://jsperf.com/hashing-strings
-export function hashString(string) {
```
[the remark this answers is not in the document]
glebm: Let's not, until we have benchmarks show that caching these would actually make programs faster.

racketscript-compiler/racketscript/compiler/runtime/core/algo/hashing.js
glebm: That would introduce a circular dependency.

Also, the two modules have different responsibilities: general low-level functions vs runtime-specific object hash code implementation.

racketscript-compiler/racketscript/compiler/runtime/core/algo/hashing.js
glebm: This is the JVM String hash. Same algorithm as before but faster implementation.

`(h << 5) - h` is `h * 31` but much faster on some browsers (Firefox).

racketscript-compiler/racketscript/compiler/runtime/core/algo/hashing.js
glebm: I'm not sure this has a name, but here is how it works.

All JavaScript numbers can be represent as 64-bit values (there is a bijection). We want to transform them to 32-bit ints, so that the distribution of the transformed values (in the 32-bit spaces) is as uniform as possible.

So we take the first half of the 64-bit value and XOR it with the second half.
This achieves the uniformity goal.

racketscript-compiler/racketscript/compiler/runtime/third-party/hash.js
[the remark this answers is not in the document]
vishesh: >  is that caching introduces a non-trivial overhead for large strings

I meant just small strings. I lazily assumed immutable people because of its popularity, must have done some measurements.

But agreed in general, and we can leave it out for now. I think premise of immutable JS people must be that if keys are strings, they are likely to be same subset of strings in an app, especially if you are modelling structured data like in JSON.

racketscript-compiler/racketscript/compiler/runtime/core/algo/hashing.js
vishesh: Cool!

racketscript-compiler/racketscript/compiler/runtime/core/algo/hashing.js
vishesh: I understood what it is doing. I just wanted to know if there is any source that talks about using  simple XOR hashing on numbers. On strings its definitely bad idea, but here we are XORing two values of 32bit width so is probably ok.

racketscript-compiler/racketscript/compiler/runtime/core/algo/hashing.js
vishesh: Now I remember why I didn't put `hashEq*` functions together with `third-party/hash.js` but in lib.js.

> Also, the two modules have different responsibilities: general low-level functions vs runtime-specific object hash code implementation.

I think its stretching the idea of modularity too far :)

We can put this in `core/raw_hashing.js` and the `hash_code.js` can be renamed as `core/hashing.js`.

glebm: > I think its stretching the idea of modularity too far

Let's agree to disagree, but for one I'm glad there would be a circular dependency issue here :)

Done

Issue #23, opened 2016-12-17
'() seems to print as ""


takikawa:
This is pretty minor, but here's an example program:

```
#lang racketscript/base

(require racketscript/htdp/universe
         racketscript/htdp/image 
         racketscript/interop
         racketscript/browser)

(displayln 3)
(displayln '(1 2 3))
(displayln '())
```

and console log (I thought the third line would be `()`):

```
Console Log: 
[log] 3 
[log] (1 2 3) 
[log]  
```

Issue #18, opened 2016-11-04
coverage: only racket tests are considered


vishesh:
When we get JS tests up (or just running fixtures), would it be possible to include that data as well?

PR #96, opened 2017-09-03
Implement print / write / display

### LINKED ISSUE #92: Printing modes
Perhaps the following functions should be on every primitive:

```typescript
toStringForWrite(): String // formerly toRawString
toStringForPrintAsExpression(): String // defaults to toStringForWrite
toString(): String // for display
```

Both should produce output readable by Racket (they mostly already do).
We could go even further if we decide to support Unicode property testing (then we can test e.g. if a character should be escaped or not).

They should return JS-native strings, as I think most of their use-cases will be IO.
We can later write optimized `toUStringFor*` versions if that proves necessary.

glebm:
Implements print / write / display with correct semantics.

For optimal performance, this implementation allows for specialization to Strings vs UStrings depending on what's best for the given output port.

For example, `console.log` wants regular Strings, but `open-output-string` wants UStrings.

Also:
* Adds `fprintf` and support for most formatting symbols.
* Changes `error` to match Racket's implementation with the default `error-value->string-handler`.

I had to split up a few modules here and there to avoid circular dependencies.

The easiest way to review this PR is by starting from the new `print*.js` files and then `primitive_print.js`.

Fixes #92 and #23.

PS: Most indentation changes (sorry) are because the original code uses tabs instead of spaces, and I've fixed it in places where I've changed most of the code anyway.


glebm:
Hi @vishesh, a few tests are failing and I have no idea why!

One test in particular is `basic/arity-case-lambda.rkt`.
This is the first example that fails:

```racket
(define lam1
  (case-lambda
    [(a b c) (* a b c)]
    [(a b) (+ a b)]))
(displayln (procedure-arity lam1))
```

It should display `'(2 3)` but displays `()` instead.
However, `procedure-arity` does return `(2 3)` and calling `display` there correctly displays `(2 3)`.

Racket defines `displayln` in [racket/private/misc.rkt](https://github.com/racket/racket/blob/62f5b2c4e4cdefa18fa36275074ff9fe376ddaf3/racket/collects/racket/private/misc.rkt#L187) like this:

```racket
  (define displayln
    (case-lambda
     [(v) (displayln v (current-output-port))]
     [(v p) 
      (unless (output-port? p)
        (raise-argument-error 'displayln "output-port?" 1 v p))
      (display v p)
      (newline p)]))
```

glebm:
My current guess is it's somehow related to this file: https://github.com/racket/racket/blob/62f5b2c4e4cdefa18fa36275074ff9fe376ddaf3/racket/collects/racket/private/norm-arity.rkt

vishesh:
@glebm I will review this later this evening and will give a shot on what's happening! Sorry about the delay.

vishesh:
@glebm I see that normalize-arity uses `sort.rkt` and sort tests are also broken so its more likely to be a problem there than with normalize arity business.

vishesh:
Can we also split this patch into multiple patches? For example, I can see one which makes new primitive for Empty, one that splits the modules, indents. Similarly the code around ports can be split by first implementing Racket stuff followed by the native string part. It will be easier to review, debug and in future revert changes with smaller patches. Thanks!

glebm:
Found the bug, it was due to my misunderstanding of the `pair?` function. Turns out an empty list is not a pair! The tests are passing now, should be good to review / merge. The coverage one timed out, probably just needs re-running (it passed locally).

> Can we also split this patch into multiple patches? For example, I can see one which makes new primitive for Empty, one that splits the modules, indents. Similarly the code around ports can be split by first implementing Racket stuff followed by the native string part. It will be easier to review, debug and in future revert changes with smaller patches. Thanks!

A primitive for Empty, the module split, and the code around ports, are all required for the new display/write/print functionality (so can't be reverted independently).


Splitting it up at this point would be an inordinate amount of effort. Sorry!

glebm:
@vishesh By the way, things like `primitive_print.js` are less than ideal (but have to be like this due to circular dependencies), this can improve massively with TypeScript which supports type-safe Mixins (so e.g. we'd have a Printable mixin instead of having it directly on Primitive).

While this PR might have some issues, I'd rather see it merged sooner rather than later (even if not perfect), as in any case this PR is an improvement over the current situation.

glebm:
The coverage warning is incorrect. Must be an issue somewhere, as `codecov` reports none of the lines in `kernel.rkt` as covered: https://codecov.io/gh/vishesh/racketscript/src/697109edfe7f77935522d3ebc67c3ce61ea61190/racketscript-compiler/racketscript/compiler/runtime/kernel.rkt

vishesh:
I'm apologize again, I'm swamped with my day work and couldn't get chance to sit though whole PR. I did restart the build last night (or this morning), and now there is one failed job.

Currently our coverage tool can only trace the compiler codepath.`kernel.rkt` being runtime that compiles to JS is going to be out of coverage from `cover` tool we use for Racket. Here's the filed ticket #18. However the percentage it calculates is on overall Racket codebase, and hence it is at 74% when it should actually be around 90+.

racketscript-compiler/racketscript/compiler/runtime/kernel.rkt
```diff
@@ -787,16 +785,25 @@
...
-(define+provide (display v [out (current-output-port)]) (#js.Kernel.display v out))
-(define+provide (print v [out (current-output-port)]) (#js.Kernel.print v out))
+(define+provide (display datum [out (current-output-port)])
+  (#js.Core.display out datum))
+(define+provide (write datum [out (current-output-port)])
+  (#js.Core.write out datum))
+(define+provide (print datum [out (current-output-port)] [quote-depth 0])
+  (#js.Core.print out datum (#js.Core.isPrintAsExpression) quote-depth))
+
+(define+provide print-as-expression
```
vishesh: This should be a parameter not `case-lambda`.

https://docs.racket-lang.org/reference/parameters.html

racketscript-compiler/racketscript/compiler/runtime/kernel.rkt
```diff
@@ -787,16 +785,25 @@
...
 ;; --------------------------------------------------------------------------
 ;; Printing
 
-(define+provide (display v [out (current-output-port)]) (#js.Kernel.display v out))
-(define+provide (print v [out (current-output-port)]) (#js.Kernel.print v out))
+(define+provide (display datum [out (current-output-port)])
+  (#js.Core.display out datum))
+(define+provide (write datum [out (current-output-port)])
+  (#js.Core.write out datum))
+(define+provide (print datum [out (current-output-port)] [quote-depth 0])
```
vishesh: If I'm not mistaken `quote-depth` is ignored. Add a TODO comment.

racketscript-compiler/racketscript/compiler/runtime/kernel.rkt
```diff
@@ -787,16 +785,25 @@
   (#js.Core.Ports.isStringPort p))
 
 (define+provide (open-output-string)
-  (#js.Core.Ports.openOutputString))
+  (#js.Core.openOutputString))
 
 (define+provide (get-output-string p)
-  (#js.Core.Ports.getOutputString p))
+  (#js.Core.getOutputString p))
```
vishesh: This should stay in `Ports` like before.

racketscript-compiler/racketscript/compiler/runtime/kernel.rkt
```diff
@@ -787,16 +785,25 @@
   (#js.Core.Ports.isStringPort p))
 
 (define+provide (open-output-string)
-  (#js.Core.Ports.openOutputString))
+  (#js.Core.openOutputString))
```
vishesh: This should stay in Ports like before.

racketscript-compiler/racketscript/compiler/runtime/kernel.js
```diff
@@ -101,7 +146,7 @@ export function random(...args) {
 // TODO: add optional equal? pred
 
 export function memv(v, lst) {
```
vishesh: These functions are no longer required actually, so you can ignore them in future. I will remove them. Also please refrain from refactoring unrelated code in a task.

racketscript-compiler/racketscript/compiler/runtime/kernel.js
```diff
@@ -9,69 +9,114 @@ export function isImmutable(v) {
...
-                    }
-                })(args[matched++]);
-            }
-        }));
-}
-
-export function listToString(lst) {
-	return Core.UString.makeMutableFromChars(
-        Core.Pair.listToArray(lst));
+export function fprintf(out, form, ...args) {
```
vishesh: More code (specially kernel primitive functions) should be written in `kernel.rkt` than `kernel.js`. Whatever is left in `kernel.js` is mostly leftover from our move to RacketScript for writing kernel.

racketscript-compiler/racketscript/compiler/runtime/kernel.js
```diff
@@ -9,69 +9,114 @@ export function isImmutable(v) {
...
+    }
 }
 
-/* --------------------------------------------------------------------------*/
-// Printing to Console
-
-export function display(v, out) {
-    /* TODO: this is still line */
-    out.write(v);
+/**
```
vishesh: `format`, `display` should be easy to move to kernel.rkt.

racketscript-compiler/racketscript/compiler/runtime/kernel.js
```diff
@@ -9,69 +9,114 @@ export function isImmutable(v) {
...
 }
 
-export function print(v, out) {
-    /* TODO: this is still line */
-    out.write(v);
+/**
+ * @param {!Core.Pair.Pair} charsList a list of chars
+ * @return {!Core.UString.MutableUString}
+ */
+export function listToString(charsList) {
```
vishesh: This should  goto UString module.

racketscript-compiler/racketscript/compiler/runtime/core/primitive.js
```diff
@@ -21,37 +19,23 @@ export class Primitive {
...
-    /**
-     * @return {!String} The deserializble string representation matching Racket's `write`.
-     */
-    toRawString() {
-        return this.toString();
-    }
-
     isImmutable() {
-        throw $.racketCoreError("Not Implemented");;
+        throw new Error('Not Implemented');
```
vishesh: Please don't replace existing way of throwing errors with new ones. I remember there was some issue with creating custom exception, but that's different issue. Stick to original code style.

racketscript-compiler/racketscript/compiler/runtime/core/ports.js
```diff
@@ -12,49 +13,74 @@ class AbstractPort extends Primitive {
...
+}
+
+/**
+ * @abstract
+ */
+export class NativeStringOutputPort extends OutputPort {
+    /**
+     * @abstract
+     */
+    consume(nativeString) {
```
vishesh: Isn't write a more appropriate name here?

racketscript-compiler/racketscript/compiler/runtime/core/primitive_print.js
```diff
@@ -0,0 +1,78 @@
+// Primitive printing extensions are defined in a separate file
```
vishesh: I really would like to avoid this. I will get back on this again.

racketscript-compiler/racketscript/compiler/runtime/core/equality.js
```diff
@@ -37,15 +37,5 @@ export function isEqual(v1, v2) {
     // Bytes are not a Primitive.
     if (Bytes.check(v1) && Bytes.check(v2)) return Bytes.eq(v1, v2);
 
-    // TODO: We should not pass arrays to this function,
```
vishesh: Why is this removed? How is byte equality done now?

racketscript-compiler/racketscript/compiler/runtime/core/print_native_string.js
```diff
@@ -0,0 +1,65 @@
+import * as Primitive from './primitive.js';
+import * as Bytes from './bytes.js';
+import * as Procedure from './procedure.js';
+import * as Ports from './ports.js';
+
+/**
+ * @param {!Ports.NativeStringOutputPort} out
+ * @param {!(Primitive.Primitive|Uint8Array|Function|String|number|boolean|undefined|null)} v
+ */
+export function displayNativeString(out, v) {
```
vishesh: This seems to have lot in common with writeNativeString.

racketscript-compiler/racketscript/compiler/runtime/core/print_ustring.js
```diff
@@ -0,0 +1,70 @@
...
+
+const TRUE_USTRING = UString.makeInternedImmutable('#t');
+const FALSE_USTRING = UString.makeInternedImmutable('#f');
+const VOID_USTRING = UString.makeInternedImmutable('#<void>');
+
+/**
+ * @param {!Ports.UStringOutputPort} out
+ * @param {!(Primitive.Primitive|Uint8Array|Function|String|number|boolean|undefined|null)} v
+ */
+export function displayUString(out, v) {
```
vishesh: This seems to have lot of similarity with write.

racketscript-compiler/racketscript/compiler/runtime/core/vector.js
```diff
@@ -9,19 +13,32 @@ class Vector extends Primitive {
...
-	    if (i != this.items.length - 1) {
-		items += " ";
-	    }
-	}
-	return "#(" + items + ")";
+    /**
+     * @param {!Ports.NativeStringOutputPort} out
+     * @param {function(Ports.NativeStringOutputPort, *)} itemFn
+     */
+    dumpToPort(out, itemFn) {
```
vishesh: writeToPort. Display/Print/Write whatever the name is, in the end you "write" to a port.

racketscript-compiler/racketscript/compiler/runtime/core/pair.js
```diff
@@ -91,23 +127,41 @@ export class Pair extends Primitive {
...
 }
 
 export function makeList(...items) {
-    let len = items.length - 1;
-    let result = Empty; /* TODO: wrap this? */
-    while (len >= 0) {
-	result = make(items[len--], result);
-    }
-    return result;
+    return items.reduceRight((result, item) => {
```
vishesh: I think I intentionally use loops to avoid the overhead of repeated function calls at places as per my best judgement. Please don't change them unless there is strong reason to, specially in an unrelated patch.

racketscript-compiler/racketscript/compiler/runtime/core/pair.js
```diff
@@ -144,25 +198,3 @@ export function listMap(lst, fn) {
...
-	if (isEmpty(lst)) {
-            return len;
-	}
-	len += 1;
-	lst = lst.cdr();
-    }
-    return len;
-}
-
-export function listLength(lst) {
```
vishesh: Where did this go?

racketscript-compiler/racketscript/compiler/runtime/core/pair.js
```diff
@@ -144,25 +198,3 @@ export function listMap(lst, fn) {
...
-    return (lst === Empty)
-	? 0
-	: lst._listLength;
-}
-
-export function isList(v) {
```
vishesh: And this?

racketscript-compiler/racketscript/compiler/runtime/core.js
```diff
@@ -1,5 +1,10 @@
 // Exports classes for creating basic data types and operation on them
 
+import './core/polyfills/string.js';
```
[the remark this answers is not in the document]
glebm: The current Babel setup doesn't do this. Won't be an issue once #107 is in or once #103 is done.

racketscript-compiler/racketscript/compiler/runtime/core.js
```diff
@@ -1,5 +1,10 @@
...
+
+import * as Primitive from './core/primitive.js';
+import './core/primitive_print.js';
```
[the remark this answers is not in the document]
glebm: No, they are different. `primitive_print` is an extension of primitive. The other `print_` files are more related to print.

racketscript-compiler/racketscript/compiler/runtime/core/equality.js
glebm: This wasn't used for byte equality. Byte equality is done in the line above.
It is removed because we no longer need to pass arrays here.

Also, `Uint8Array` is not an `Array`, despite the name.

racketscript-compiler/racketscript/compiler/runtime/core/pair.js
glebm: It's now in kernel.rkt as: 

```racket
(define+provide (length v) #js.v.length)
```

racketscript-compiler/racketscript/compiler/runtime/core/pair.js
glebm: Same

racketscript-compiler/racketscript/compiler/runtime/core/ports.js
glebm: `consume` and `produce` are names used by Racket in the [documentation](https://docs.racket-lang.org/reference/ports.html) and the [code](https://github.com/racket/racket/blob/1309cf164930cca4dfed4956db8ff1048ded3d7f/racket/collects/racket/port.rkt).

racketscript-compiler/racketscript/compiler/runtime/core/primitive.js
glebm: 1. This is not a Racket error, it's an internal error that should never happen, really more of an assertion.
2. Now that Errors are more like real Racket errors, `primitive.js` can no longer depend on them to avoid circular dep.

racketscript-compiler/racketscript/compiler/runtime/core/primitive_print.js
glebm: I know exactly how to avoid this, a `Printable` mixin.
However, let's leave it is a TODO, it doesn't have to be done right now.

racketscript-compiler/racketscript/compiler/runtime/core/print_native_string.js
glebm: It is, but we can always factor out the common stuff later.

racketscript-compiler/racketscript/compiler/runtime/core/print_ustring.js
glebm: Same as above, can leave a TODO to factor out the common stuff later.

racketscript-compiler/racketscript/compiler/runtime/core/vector.js
glebm: Done

racketscript-compiler/racketscript/compiler/runtime/kernel.js
glebm: Well `fprintf` would be very difficult to read if written in Racket DSL.

racketscript-compiler/racketscript/compiler/runtime/kernel.js
glebm: `format` I think will be harder to follow in kernel.rkt (especially format as it has a spread `...`).
As for `display`, etc, these functions are used internally by the JavaScript runtime, so should stay in JS.

racketscript-compiler/racketscript/compiler/runtime/kernel.js
glebm: Having this in UString may introduce circular dep in the future if Pair gets a UString-optimized versions of d/p/w.

racketscript-compiler/racketscript/compiler/runtime/kernel.js
glebm: OK, got it. Won't revert it now though if they'll be removed anyway (and the `!== false` changes are actually bugfixes).

racketscript-compiler/racketscript/compiler/runtime/kernel.rkt
glebm: In the current design that would be a circular dep (as explained in the comment on top of `ports_ustring.js`).
We can rename the `Core` exports later (in a different PR).

racketscript-compiler/racketscript/compiler/runtime/kernel.rkt
glebm: ditto

racketscript-compiler/racketscript/compiler/runtime/kernel.rkt
glebm: It's not ignored, it's supported.

racketscript-compiler/racketscript/compiler/runtime/kernel.rkt
glebm: I don't understand parameters yet. Added a TODO, perhaps you can fix it in a follow-up PR?

glebm:
> There is too much criss-crossing across modules, rather than more layered design and probably that's why we ended up with cyclic dependencies.

The correct more-layered design is quite obvious. Introduce a `PrintablePrimitive` class (or a `Printable` mixin/interface) to avoid patching the `Primitive.prototype` directly. However, doing this without TypeScript and interfaces is error-prone and hard-to-follow (`UString` is printable, but `PrintablePrimitive` also uses `UString` in its interface), so I've left it for the future.

> I'm sorry, but I can't accept in its current state. It needs to be split across logically separate patches! There are too many unrelated changes (which don't even matter to test this change) and I can't constructively review and approve it with confidence.

The changes are not as unrelated as a cursory look might show. Please see my responses to your comments.

 > `write` and `display` are clearly too similar and is an indicator that it can be abstracted out. Having specialized implementation for native and ustring is fine but it seems like it can be abstracted out for its client in better ways (composition). 

The client only needs to implement the native string version, the ustring one is already optional!

glebm:
@vishesh I'm going to look into making a `Printable` interface / mixin.

vishesh:
Please split patch before anything. I will get back on other comments later.

glebm:
Hey, I've removed most of the circular dependency workarounds by introducing `PrintablePrimitive` and `MiniNativeOutputStringPort`. They are small separate commits above, so should be easy to follow.

> Please split patch before anything. I will get back on other comments later.

Already did the above by the time I saw this. I'll split at least the `Empty` change out later tonight.

PR #98, opened 2017-09-03
Fixes hdtp to pass native strings to JS

glebm:
hdtp API still accepts Racket strings so that doesn't change, but internally it now has to pass native strings to JS APIs.

### DIFF
--- a/racketscript-extras/racketscript/htdp/image.rkt
+++ b/racketscript-extras/racketscript/htdp/image.rkt
@@ -82,10 +82,10 @@
               #,(when (syntax-e #'pen)
                   #`(cond
                       [(or (equal? mode 'outline) (equal? mode "outline"))
-                       (:= #js.ctx.strokeStyle (->web-color pen))
+                       (:= #js.ctx.strokeStyle (js-string (->web-color pen)))
                        (#js.ctx.stroke)]
                       [(or (equal? mode 'solid) (equal? mode "solid"))
-                       (:= #js.ctx.fillStyle (->web-color pen))
+                       (:= #js.ctx.fillStyle (js-string (->web-color pen)))
                        (#js.ctx.fill)]))
               (#js.ctx.closePath))]))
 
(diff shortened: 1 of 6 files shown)


vishesh:
Thanks a lot for this! Did you try some examples from here https://github.com/vishesh/racketscript-playground/tree/master/examples ?

glebm:
I've sent a PR to the playground as well. The only one I've tried and fixed for now is colored-carpet (only the jQuery calls needed fixing).

racketscript-extras/racketscript/htdp/image.rkt
```diff
@@ -166,11 +167,11 @@
    (λ (ctx x y)
      (with-origin ctx [x y]
        (set-object! ctx
-                    [font          #js*.this.font]
-                    [textAlign     "center"]
-                    [textBaseline  "middle"]
-                    [fillStyle     #js*.this.color])
-       (#js.ctx.fillText #js*.this.text 0 0)))])
+                    [font          (js-string #js*.this.font)]
```
vishesh: Can we convert this (and other js-string in render) when we are updating the object?

vishesh:
Missed sending this:

@glebm Can you please do a quick check if games are running fine? Its a quick indicator if things are running fine. If something doesn't work, its a bug with HTDP or RacketScript. Since you fixed Playground, you can just run other games from there. Changes looks good, and hopefully nothing should fail! :)

racketscript-extras/racketscript/htdp/image.rkt
glebm: While we can, I think it is simpler Racket Strings internally and JS strings only when passing to JS APIs. This way Racket code can be used to manipulate strings, and it is a clear decision whether something is a Racket string or a JS string.

glebm:
Yeah there are some issues with games, hopefully easy to fix, I'll look into it

glebm:
@vishesh OK, done, should be good to merge.

All the games work as before. Flappy bird still has the accidental initialization issue but it is present on http://rapture.twistedplane.com:8080/#example/flappy-birds as well

vishesh:
Thanks!

PR #102, opened 2017-09-05
A `make tscheck` task to typecheck the JS runtime

glebm:
Currently fails to run due to some issues with the type signatures but they're fixed in #96.


vishesh:
Thanks

PR #105, opened 2017-09-05
add hash predicates; more hash tests

stchang:


stchang:
Btw, does hamt support weak maps?

glebm:
Hamt doesn't seem to support weak maps. Racket doesn't support immutable weak maps though.

At least the `weakhash-eq` can currently be implemented simply keeping a [WeakMap](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap) for objects and a regular [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) for JS primitives.

`weakhash-eqv` and `weakhash-eqv` are a bit trickier.

vishesh:
Thanks @stchang. I think we don't have to worry about weak-hashequal/eqv. `weak-eq` is easy using WeakSet.

stchang:
Sure I can do it. Though I may not get to it until next week. I still plan to implement the `iterate-` functions too.

vishesh:
Awesome! I will merge this meanwhile.

PR #107, opened 2017-09-06
Switch to only supporting Node 8

### LINKED ISSUE #101: Drop NodeJS < 8
This will let us care less about polyfills.

For example, `String#padStart` is supported in all browsers, but in NodeJS only since v8.2.1+ http://node.green/#ES2017-features-String-padding.

As this is an experimental project, we shouldn't care about supporting older versions of Node / Racket / etc.
Once this project is no longer experimental, all the current versions will have become old.

glebm:
Ahead of its becoming an LTS release a few months from now.
Fixes #101

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,8 @@
 language: node_js
 
 node_js:
-  - "5"
-  - "6"
-  #- "node"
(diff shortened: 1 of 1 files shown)


glebm:
The tests now run twice as fast, as we no longer hit the parallelization limit :smile:

PR #108, opened 2017-09-06
List improvements

glebm:
* Adds a singleton class for Empty.
* Makes `list?` O(1) instead of O(n).
* Makes `length` checked.
* Makes `immutable?` work on lists.
* Makes the `_listLength` field always a number so it's more JIT-friendly.
* Avoids the expensive `instanceof` checks in `Pair.check` and `MPair.check`.
* Adds hash code caching for `Pair` and `MPair`.
* Fixes the `listFind` predicate check.
* Minor improvements to `makeList` and `listFromArray`.
* Removes the unused `_listLength` function.
* Removes the now unnecessary Array equality and hashing from `isEqual` and `hashForEqual`.
* Removes `cons?` and `empty?` from kernel as they are not a part of the Kernel in racket.
* Adds more tests.

@vishes Hoping for a quick turnaround on this so I can proceed with extracting the rest of the mega-PR.

### DIFF
--- a/racketscript-compiler/racketscript/compiler/runtime/core/equality.js
+++ b/racketscript-compiler/racketscript/compiler/runtime/core/equality.js
@@ -37,15 +37,5 @@ export function isEqual(v1, v2) {
     // Bytes are not a Primitive.
     if (Bytes.check(v1) && Bytes.check(v2)) return Bytes.eq(v1, v2);
 
-    // TODO: We should not pass arrays to this function,
-    //   but currently we sometimes do, e.g. the empty list.
-    if (Array.isArray(v1) && Array.isArray(v2) && v1.length === v2.length) {
-        const n = a.length;
-        for (let i = 0; i < n; i++) {
-            if (!isEqual(a[i], b[i])) return false;
-        }
-        return true;
-    }
-
     return false;
 }
--- a/racketscript-compiler/racketscript/compiler/runtime/core/hash.js
+++ b/racketscript-compiler/racketscript/compiler/runtime/core/hash.js
@@ -135,7 +135,7 @@ export function makeEqualFromAssocs(assocs, mutable) {
 }
 
 export function map(hash, proc) {
-    let result = Pair.Empty;
+    let result = Pair.EMPTY;
     hash._h.forEach((value, key) => {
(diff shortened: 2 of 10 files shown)


stchang:
Maybe add tests using `cons?` and `empty?` to `tests/racket-core/list.rkt`?

Otherwise looks ok to me.

glebm:
@stchang I thought these tests were meant to match https://github.com/racket/racket/blob/master/pkgs/racket-test-core/tests/racket/list.rktl? That doesn't have tests for `cons?` and `empty?`.

vishesh:
LG.

stchang:
@glebm Yes, you're right. Perhaps it's better to add a new test file then?

glebm:
Well, we're not testing the rest of non-core Racket.

If our implementation of the Kernel is correct, as verified by the kernel tests, we don't need to test the non-kernel code.

glebm:
Anyway, there were no tests for these functions before, and now the functions no longer exist. The addition of these tests should not block this PR.

Can we merge this for now, independently of what you can decide your policy is on testing non-kernel code?

stchang:
> If our implementation of the Kernel is correct, as verified by the kernel tests, we don't need to test the non-kernel code.

Racketscript semantics are different from Racket so we cannot assume this.

And just because Racket tests are incomplete does not mean Racketscript must be.

> now the functions no longer exist

That's exactly my concern. Existing code may use these functions but after this commit I do not know if they still work.

Anyhow, it's a minor point here so I'm ok with merging either way

glebm:
> That's exactly my concern. Existing code may use these functions but after this commit I do not know if they still work.

As far as I can tell, existing code already used the `list.rkt` definitions as they override `kernel.rkt`.

By the way, the definitions in `list.rkt` are just aliases:

https://github.com/racket/racket/blob/0d5948683923319a959db003fe3d2ab5c54a4b75/racket/collects/racket/list.rkt#L114-L115

Once you merge this, if you want to add the tests, here are some:

```racket
(empty? '())
(empty? '(1))
(empty? #t)

(cons? '())
(cons? '(1 2))
(cons? (cons 1 2))
(cons? (cons 1 (cons 2 3)))
```

PR #111, opened 2017-09-07
Add contribution guidelines

vishesh:
Can be better read here: https://github.com/vishesh/racketscript/blob/issue/contribution-guidelines/CONTRIBUTING.md

PR #114, opened 2017-09-07
.eslintrc and `make eslint-fix`

glebm:
As per #111, this project now uses the AirBnb JavaScript style guide but with 4 spaces for indentation.

This configures the project using the official AirBnb eslint configuration and adds an eslint-fix task that can fix some of the issues.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,11 @@
+{
+    "extends": "airbnb-base",
+    "parserOptions": {
+        "ecmaVersion": 2017,
+        "sourceType": "module"
+    },
+    "rules": {
+        "indent": ["error", 4],
+        "import/extensions": ["error", "always"]
+    }
+}
--- a/Makefile
+++ b/Makefile
@@ -9,9 +9,10 @@
 .PHONY: build setup setup-extra clean
 .PHONY: test unit-test integration-test test
 .PHONY: coverage coverage-unit-test
-.PHONY: eslint jslint tscheck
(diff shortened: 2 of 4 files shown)


glebm:
> Our of curiosity apart from string padding what else do we use from es2017?

As far as I know:

es2017: TextEncoder/Decoder, String padding
es2016: Modules (`import/export` spec)
es2015 (aka es6): Uint8Array constructor, ArrayBuffer

> I think we can remove jshint jslint completely from this Makefile. No one uses it.
> I can't see from my cellphone, but since eslint is installed here maybe we should remove eslint and jshint the 'install -g` business from setup-extra and invoke npm intsall from that target.

+1, I'll send a separate PR later on with these two changes.

> Also what do you think about doing 'npm install --save-dev' instead of maintaining package.json? Sure package.json is better, but it's presence in root makes it look like a NPM package and hence a self-hosted compiler which we are not. I may be overthinking this.

`--save-dev` also goes to `package.json`. I mean, this package does have NPM dev dependencies so it makes sense to have them specified. Note that we also check in `package-lock.json`, which is uncommon for packages.

You can move it under `/racketscript-compiler/racketscript/compiler/runtime/` if you're concerned though, and simply adjust the Makefile.

vishesh:
Yes except package.json won't be part of repo in savedev case and generated. Anyway it's just trivial cosmetic thing. I will merge this.

TypeScript also I believe should be a dev dependency.

PR #115, opened 2017-09-07
Remove text_transcoder.js polyfill

glebm:
Use the built-in TextDecoder/Encoder instead.

It is still experimental but will likely be stabilized in the future with few changes if any (WRT UTF-8 encodings, the other encodings are still in flux but we don't use them).


vishesh:
What happens to browser targets ?

glebm:
@vishesh No special handling, the `typeof` check will succeed if the browser supports it or if a polyfill is present.

vishesh:
Great. Thanks!

PR #116, opened 2017-09-07
Move procedure-related functions to procedure.js

glebm:
Extracted from the mega-PR


vishesh:
Thanks.

stchang:
Should we change the display functions to use the new procedure `toString`?

glebm:
@stchang, no need for now as I'll send a PR soon that does the full display/write/print implementation

stchang:
Great thanks.

PR #118, opened 2017-09-09
Makefile fixes

glebm:
Correctly specifies dependencies between targets.
Removes no-longer necessary steps from the readme.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,91 +1,74 @@
-##   mmmmm                  m
-##   #   "#  mmm   mmmm   mm#mm  m   m   m mm   mmm
-##   #mmmm" "   #  #" "#    #    #   #   #"  " #"  #
-##   #   "m m"""#  #   #    #    #   #   #     #""""
-##   #    " "mm"#  ##m#"    "mm  "mm"#   #     "#mm"
-##               #
-##               "
+.PHONY: build setup clean \
+eslint eslint-fix tscheck \
+test unit-test integration-test coverage coverage-unit-test
 
-.PHONY: build setup setup-extra clean
-.PHONY: test unit-test integration-test test
-.PHONY: coverage coverage-unit-test
-.PHONY: eslint eslint-fix jshint tscheck
+SETUP_BIN_DIR=$(shell racket -e "(require setup/dirs) (display (find-user-console-bin-dir))")
+SETUP_PKGS_DIR=$(shell racket -e "(require setup/dirs) (display (find-user-pkgs-dir))")
+TSC=node_modules/.bin/tsc
+ESLINT=node_modules/.bin/eslint
 
-TSC=node_modules/typescript/bin/tsc
-ESLINT=node_modules/eslint/bin/eslint.js
+## Compiler recipes
 
-## Compile recipes
+build: setup
+	raco make -j 2 -v racketscript-compiler/racketscript/compiler/main.rkt \
+	  tests/fixture.rkt
 
-build:
-	@echo "     RACKETSCRIPT BUILD      "
-	@echo "++++++++++++++++++++++++++++++"
-	@echo
-	@echo "NOTE: Make sure NodeJS and NPM are installed and"
-	@echo "      exists in PATH. See \`make setup\`"
-	@echo
-	raco make -v racketscript-compiler/racketscript/compiler/main.rkt
-	raco make -v tests/fixture.rkt
-
-setup:
+$(SETUP_BIN_DIR)/racks:
 	raco pkg install --auto -t dir racketscript-compiler/ || \
-	    raco pkg update --link racketscript-compiler/
+	  raco pkg update --link racketscript-compiler/
 	raco pkg install --auto -t dir racketscript-extras/ || \
-		    raco pkg update --link racketscript-extras/
+	  raco pkg update --link racketscript-extras/
+$(SETUP_PKGS_DIR)/cover:
+	raco pkg install --auto cover || true
+$(SETUP_PKGS_DIR)/glob:
+	raco pkg install --auto glob || true
+setup: | $(SETUP_BIN_DIR)/racks $(SETUP_PKGS_DIR)/cover $(SETUP_PKGS_DIR)/glob
 
-setup-extra:
-	npm install -g traceur js-beautify eslint jshint gulp
-	raco pkg install --auto cover glob
-
-clean:
-	raco pkg remove racketscript-extras
-	raco pkg remove racketscript-compiler
-
-## Coverage recipes
-
-coverage-unit-test:
-	@echo " RACKETSCRIPT COVERAGE UNIT-TEST    "
-	@echo "++++++++++++++++++++++++++++++++++"
-	raco cover -d ./coverage/unit racketscript-compiler/racketscript/
-
-coverage:
-	@echo "    RACKETSCRIPT COVERAGE    "
-	@echo "++++++++++++++++++++++++"
-	COVERAGE_MODE=1 raco cover -d ./coverage/all -b racketscript-compiler \
-	    tests/fixture.rkt
+clean: clean-compiler
+	raco pkg remove racketscript-extras || true
+	raco pkg remove racketscript-compiler || true
+	rm -rf test/compiled/ coverage/ tmp/ js-build/
+	bash -c 'rm -rf racketscript-compiler/racketscript/compiler/**/*.{dep,zo}'
 
 ## JavaScript
 
 eslint: | node_modules
-	@echo "    RACKETSCRIPT RUNTIME LINT    "
-	@echo "++++++++++++++++++++++++++++"
-	$(ESLINT) ./racketscript-compiler/racketscript/compiler/runtime/ || true
+	$(ESLINT) ./racketscript-compiler/racketscript/compiler/runtime/
 
 eslint-fix: | node_modules
 	$(ESLINT) --fix ./racketscript-compiler/racketscript/compiler/runtime/
-
-jshint:
-	@echo "    RACKETSCRIPT RUNTIME LINT    "
-	@echo "++++++++++++++++++++++++++++"
-	jshint ./racketscript-compiler/racketscript/compiler/runtime/ || true
 
 # Typecheck JavaScript
 tscheck: node_modules
 	$(TSC) --noEmit --allowJs --checkJs --strict --lib es2017 --target es2017 \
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -1,75 +1,51 @@
...
-	@echo "++++++++++++++++++++++++++++++++++"
-	raco cover -d ./coverage/unit racketscript-compiler/racketscript/
+## Compiler recipes
 
-coverage:
-	@echo "    RACKETSCRIPT COVERAGE    "
-	@echo "++++++++++++++++++++++++"
-	COVERAGE_MODE=1 raco cover -d ./coverage/all -b racketscript-compiler \
-	    tests/fixture.rkt
+build: setup
```
vishesh: Please remove `setup` from this as it can get slightly annoying to keep reinstalling package after each small change. When testing compiler changes you just want to build the bytecode and expect setup to be done already.

Also other targets which uses build (tests ..) will do setup again and again.

Makefile
```diff
@@ -1,75 +1,51 @@
...
+	raco make -j 2 -v racketscript-compiler/racketscript/compiler/main.rkt \
+	  tests/fixture.rkt
+
+$(SETUP_BIN_DIR)/racks: $(call rwildcard,,*.rkt)
```
vishesh: What role `$(call ..)` plays in this target?

README.md
```diff
@@ -93,19 +93,19 @@ Traceur using NPM. Babel is recommended for writing NodeJS programs.
```
vishesh: Can we remove this change? This may give merge conflict with the contribution guidelines. HACKING is already gone in that branch, and I will remove trailing white-spaces.

Makefile
```diff
@@ -1,75 +1,51 @@
...
 	$(ESLINT) --fix ./racketscript-compiler/racketscript/compiler/runtime/
 
-jshint:
-	@echo "    RACKETSCRIPT RUNTIME LINT    "
-	@echo "++++++++++++++++++++++++++++"
-	jshint ./racketscript-compiler/racketscript/compiler/runtime/ || true
-
 # Typecheck JavaScript
-tscheck: node_modules
+tscheck: | node_modules
```
vishesh: This comment applies to other targets like eslint .... Since these targets will be used by contributors, we can assume they have done the `setup-extras` step by following the contribution guidelines. So we can avoid the `npm install` lag each time for our convenience

Makefile
glebm: I was under the impression that changes to `.rkt` files require an install to work? Is that not so?

Makefile
glebm: `call` is the built-in Makefile function that calls a user-defined function.
We specify the pre-requisites of `racks` so that `setup` only runs when one of the `.rkt` files have changed.

Makefile
glebm: No. `node_modules` specifies `package.json` as it's pre-requisite. If neither `package.json` nor `node_modules/` have changed, `npm install won't run.

README.md
glebm: Reverted

Makefile
vishesh: No. The package is linked to current directory. You just need to do a `raco make`.

Makefile
vishesh: `raco make` will already take care of that, so you don't have to worry about.

Makefile
vishesh: Seems like it does. I'm running `make tscheck` twice and it takes around 5 seconds to get this

```
± % make tscheck                                                                                                                                                 
npm install                                                                                                                                                                               
npm WARN racketscript No repository field.                                                                                                                          
npm WARN racketscript No license field.                                                                                                                             
                                                                                                                                                                   
up to date in 0.86s                                                                                                                             
```

glebm: @vishesh Nice catch, that was a bug.

Your `node_modules` was older than `package.json`.

```
stat -c %Y node_modules
stat -c %Y package.json
```

This can happen if there was a change to `package.json` that didn't result in `node_modules` getting updated (such as `typescript` moving from `dependencies` to `devDependencies`).

I've added ` && touch node_modules` to the rule to avoid this.

Makefile
glebm: This was to avoid needless `raco pkg install`, not `raco make`.

Makefile
glebm: I'll see if I can make it run only once.

glebm: Done, now `setup` only runs once.

vishesh: I really don't think its necessary to do more than `raco make`.

Developers would have already run `setup-extra` which should take care of installing this all. No need to add complexity where its not needed.

Makefile
glebm: Removed the `**/*.rkt` pre-requisites so that `setup` only runs once.

vishesh: I think we don't need this. If I'm doing setup I really want everything reinstalled for some reason. Keep it simple.

Makefile
vishesh: Like I said in other comment, please don't worry about that. `setup-extra` should handle whatever installation stuff you have to. Let's keep thing whole thing simple and stupid.

glebm:
> I really don't think we should do installation of node modules for eslint, tscheck ... First thing developer must to is run make setup-extra which should install whatever is necessary. Overall, keep it simple and straight.

1. We don't run `npm install` if `node_modules` are already installed. If they are not installed and a task requires them, only then do they get installed.
2. There is no longer a `setup-extra` step at all in this Makefile. Removing it makes the whole thing simpler. Having an extra step instead of letting `make` take care of it is not what I'd consider simple.

If you think that install for people who just want to install the package and not hack on it should happen via `make setup` (I think it should just be `raco pkg install`, is that right?), we can add the setup-extra back with a dependency.

Makefile
glebm: @vishesh Bringing your attention again to this comment, seems like you've missed it.
TLDR: That was a bug and is now fixed.

vishesh:
Here's how it is:

1. You are a user and you just want to use it. You do `make setup`. You DON'T care about anything else. 
2. You are a hacker and want to fiddle with things. You do `make setup-extra` too. It will install whatever NPM and Racket developer dependencies required. `make-setup-extra` can simply call `setup-extra-racket` and `setup-node-modules` whatever. Personally, I like to just do `make setup-extra` whenever I want to update my environment or install in a fresh machine.

By this point, you have your developer environment setup. You have all your tools installed (eslint, tscheck, cover ...) from setup-extra. And you will run `make build`, `make eslint` etc . several times. I don't understand the reason why you want to do all these checking and installation again for eslint and all targer. Sure its fancier, but more code with no real benefit, because I repeat, you already have done setup-extra!

>> raco pkg install

There are people who keep latest repos just to keep up-to-date with latest changes. My deployment scripit in RacketScript playgound pull latest changes and simply does a `raco setup`.

Overall, I found this overall workflow good enough to work with, and don't feel the need for changes there.

Makefile
vishesh: I guess this is exactly why I emphasize on making things simple. Simply do a `npm install` or `raco make` than checking if XYZ changed. Let NPM or Raco handle whatever it has to.

glebm:
OK, I've updated the PR, so `make setup` now only runs the non-development setup.

With this change, you no longer need to worry about `setup-extra`, at all.
This is much simpler than before.

> Sure its fancier, but more code with no real benefit,

The real benefit is not having to understand what setup-extra does and **when I need to re-run it**.
The previous setup only seemed simple to you because you knew exactly what `setup-extra` did.

Makefile
```diff
@@ -1,91 +1,76 @@
...
+$(SETUP_PKGS_DIR)/glob:
+	raco pkg install --auto glob || true
+setup-extras: | setup $(SETUP_PKGS_DIR)/cover $(SETUP_PKGS_DIR)/glob
 
-coverage:
-	@echo "    RACKETSCRIPT COVERAGE    "
-	@echo "++++++++++++++++++++++++"
-	COVERAGE_MODE=1 raco cover -d ./coverage/all -b racketscript-compiler \
-	    tests/fixture.rkt
+build: setup-extras
```
vishesh: I think I mentioned earlier to remove setup-extras from here.

Makefile
```diff
@@ -1,91 +1,76 @@
...
-	raco pkg remove racketscript-compiler
+## Compiler recipes
 
-## Coverage recipes
+$(SETUP_BIN_DIR)/racks:
+	raco pkg install --auto -t dir racketscript-compiler/ || \
+	  raco pkg update --link racketscript-compiler/
+	raco pkg install --auto -t dir racketscript-extras/ || \
+	  raco pkg update --link racketscript-extras/
+setup: | $(SETUP_BIN_DIR)/racks
```
vishesh: I want setup to run regardless it is installed or not. That's why there is a `raco update`. Right now:

```
± % make setup
make: Nothing to be done for 'setup'.
```

I think what I like to have is to Makefile act just like a function to run some commands (that is why I added PHONY) like I would otherwise do on bash. I really have no interest in maintaining Makefiles more than necessary and keep it simple. And that is why I'm asking you to remove node_modules target from other targets, Rather than checking if package.json file changed, just run npm install whatever.

So far this way of working with Makefile has served all our use-cases well, and there is no need to change something that is working just fine, let alone make it do more. Whatever confusion you have around `setup-extras` is just your experience so far. So lets not fix a problem that not yet exist. I would be more open to this if there is more shoutout for a change.

On side note there are other changes I like, for example building before running unit-test which is almost always relevant.

glebm:
My experience highlighted the problem. It's just me but unfortunately, there are no other contributors for now who are not fully familiar with the codebase, to demonstrate that the previous setup is less than ideal.

Closing this PR as I have a rather different idea of what it should look like (or what a good development experience is).

glebm:
> I think what I like to have is to Makefile act just like a function to run some commands

This is a really weird approach to makefiles. I've never seen this before in over a decade of dealing with makefiles.

You can always force a makefile to run the command by running `clean` first.
If you want the commands to always run, don't use makefiles, just use bash scripts.

Using a Makefile like this is kinda like using an plane as a car, because the plane happens to have wheels and you don't really like flying...

vishesh:
Regarding the patch, there are certainly improvements, I've already given my opinion on most parts.

For rest, there can't be a consensus between two if opinions and priorities are different, and I think our time is better spent discussing other more important issues on hand.

PR #120, opened 2017-09-09
Update .eslintrc `make eslint-fix`

glebm:
.eslintrc changes:

* Disallow dangling commas as they do not always work in NodeJS.
* Allow `~` by default as it is not likely to be a typo.
* Allow leading underscores by default when called via `this`, e.g.
  `this._private_prop`.
* Allow use-before-define for functions. It is safe as function
  declarations are hoisted, and we do it throughout the codebase.
* Allow generator loops (`for ... of`), as we do not target older
  browsers that do not support them (AirBnB bans them by default because
  the polyfill is too heavy).

There are still some lint errors but they will need to be fixed /
ignored manully.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -6,6 +6,32 @@
     },
     "rules": {
         "indent": ["error", 4],
-        "import/extensions": ["error", "always"]
+        "import/extensions": ["error", "always"],
+        "comma-dangle": ["error", "never"],
+        "no-bitwise": ["error", { "allow": ["~"] }],
+        "no-underscore-dangle": ["error", { "allowAfterThis": true }],
+        "no-use-before-define": [
+            "error",
+            {
+                "functions": false,
+                "classes": true,
+                "variables": true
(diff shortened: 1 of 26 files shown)


glebm:
@vishesh Please review the changes as the AirBnB style guide does make some unusual choices by default (like not indenting a `switch/case`).

vishesh:
Thanks @glebm!

Regarding `switch` statement, its actually pretty common in C/C++/Java  world. Maybe that's how they picked it up. What else you find unusual?

BTW are spread arguments now natively supported by browsers and node?

glebm:
@vishesh Yes, spread arguments are supported everywhere (http://kangax.github.io/compat-table/es6/#test-spread_(...)_operator). So is most of ES6+, and even ES2017, most browsers are evergreen nowadays.

I didn't find anything else to be unusual, just seeing that `switch` was enough for me to question the rest.

PR #121, opened 2017-09-09
internedMake: Use a Map instead of Object

glebm:
Using objects as maps is dangerous and leads to hard-to-debug cases.
E.g. consider what happens if someone puts `"hasOwnProperty"` as a key.

Actual `Map`s are also slightly faster in most modern browsers.


vishesh:
Good catch. I believe we made it for symbol, strings but could have been easily misunderstood for others.

PR #124, opened 2017-09-10
Fix TextEncoder/Decoder in browsers

glebm:
Fixes a bug introduced in #115.
`var` got hoisted even if the `if` check failed, resulting in undefined TextEncoder/Decoder.

Also fixes a missing import in box.js.


vishesh:
Thanks!

PR #125, opened 2017-09-10
Fix parameterize to false'ish values.

glebm:
The previous version made it impossible to parameterize to false, 0, etc.

Also adds handling for `getCurrentParametrization()` returning false.

### DIFF
--- a/racketscript-compiler/racketscript/compiler/runtime/paramz.js
+++ b/racketscript-compiler/racketscript/compiler/runtime/paramz.js
@@ -37,7 +37,8 @@ export function makeParameter(initValue) {
     const param = function (maybeSetVal) {
         // Get current value box from parameterization. The function
         // `param` that we result is the key.
-        let pv = getCurrentParameterization().get(param, false) ||
+        const current = getCurrentParameterization();
+        let pv = (current && current.get(param, false)) ||
             __top.get(param, false);
         // Create entry in __top if its a mutation.
         if (!pv && maybeSetVal !== undefined) {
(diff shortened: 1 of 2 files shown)


tests/wcm/parameter-set.rkt
```diff
@@ -16,4 +16,9 @@
     (set! p2 p2*)
     (displayln (list (p1) (p2)))))
 
+(define bool-param (make-parameter #t))
+(displayln (bool-param))
+(parameterize ([bool-param #f])
+  (displayln (bool-param)))
```
vishesh: can we also add:

```
(parameterize ([bool-param 0])
  (displayln (bool-param))
```

racketscript-compiler/racketscript/compiler/runtime/paramz.js
```diff
@@ -37,7 +37,8 @@ export function makeParameter(initValue) {
     const param = function (maybeSetVal) {
         // Get current value box from parameterization. The function
         // `param` that we result is the key.
-        let pv = getCurrentParameterization().get(param, false) ||
+        const current = getCurrentParameterization();
+        let pv = (current && current.get(param, false)) ||
```
vishesh: There must always be a parameterization in context and we should assume that here. Otherwise it can be a bug. Are we getting falsy values from `getCurrentParameterization` in any case?

glebm: It happens when I run the test suite with #126

vishesh: The fix looks good. Please let me take a look at `getCurrentParameterization()` though.

tests/wcm/parameter-set.rkt
glebm: Done

vishesh:
Don't have time to look at `getCurrentParameterization()` stuff. I will merge this, and create a ticket for that so I don't forget what is happening there.

Thanks again!

Issue #93, opened 2017-08-30
Implement the Racket number pyramid


vishesh:
Racket numbers are exact, in-exact, complex and can be arbitrarily large as opposed to 64-bit floats in JavaScript. 

- Whalesong and Pyret use Danny Yoo's [`js-numbers`](https://github.com/brownplt/pyret-lang/blob/master/src/js/base/js-numbers.js) library. IIRC the library itself models the number pyramid, and hence can be used directly.

vishesh:
@glebm If you are interested you can take this. A lot of people ask about this.

avelino:
PR merged, I think you can close this issue

vishesh:
@avelino: The PR references this issue, but not fixes it.

samth:
1. Danny Yoo implemented the full numeric tower, it's part of the wescheme implementation here: https://github.com/bootstrapworld/wescheme2012/blob/master/war-src/js/js-runtime/js-numbers.js
2. Modern browsers now have bignums, which would make that code a lot faster.

vishesh:
Native `BigInt` is great news! Thanks for update!!

sorawee:
It still doesn't support fractions though, so we probably want to use a library for that instead of reinventing the wheel. https://github.com/infusion/Fraction.js/blob/master/bigfraction.js looks promising to me (it uses `BigInt` under the hood -- most other libraries implement their own `BigInt` too).

PR #127, opened 2017-09-11
Fix broken HTDP libs

vishesh:
HTDP got broken after changes after improvements to identifier normalization improvements (expect more breaks :( in future around code that use FFI).

I'm putting this on CR to get some feedback on how to handle `this` stuff in JavaScript in future from RacketScript. Its not a safe thing to use, and particularly when you are using it in a callback. What this patch does is to extend[1] lambda's bind `this` with a variable, which will make it safe to use in callbacks and so.

Should we encourage something like this in our FFI  whenever `this` is involved? There will always be raw low-level `this` available if one actually needs that.

[1] https://github.com/vishesh/racketscript/compare/develop...task/fix-htdp-libs?expand=1#diff-6a9c6fd3ed8118a4107bf70798eb983bR47

@stchang @glebm

### DIFF
--- a/racketscript-extras/racketscript/htdp/image.rkt
+++ b/racketscript-extras/racketscript/htdp/image.rkt
@@ -124,7 +124,8 @@
 
 (define-proto EmptyScene
   (λ (width height borders?)
-    (set-object! #js*.this
+    #:with-this this
+    (set-object! this
                  [type      "empty-scene"]
                  [width     width]
                  [height    height]
@@ -135,7 +136,8 @@
 
 (define-proto Text
   (λ (text size color face family style weight underline?)
-    (set-object! #js*.this
+    #:with-this this
+    (set-object! this
                  [type       "text"]
                  [text       text]
                  [size       size]
@@ -145,47 +147,51 @@
                  [style      style]
                  [weight     weight]
                  [underline  underline?])
-    (#js*.this._updateMetrics))
+    (#js.this._updateMetrics))
   [_updateMetrics
    (λ ()
+     #:with-this this
      (define font (string-append
-                     #js*.this.weight " "
-                     #js*.this.style " "
-                     (number->string #js*.this.size) "px "
-                     #js*.this.face " "
-                     #js*.this.family))
+                     #js.this.weight " "
+                     #js.this.style " "
+                     (number->string #js.this.size) "px "
+                     #js.this.face " "
+                     #js.this.family))
 
      (:= ($ *invisible-canvas-context* 'font) (js-string font))
      (define metrics ($> *invisible-canvas-context*
-                         (measureText (js-string #js*.this.text))))
+                         (measureText (js-string #js.this.text))))
 
-     (set-object! #js*.this
+     (set-object! this
                   [font    font]
                   [width   #js.metrics.width]
-                  [height  #js*.this.size]))]
+                  [height  #js.this.size]))]
   [render
    (λ (ctx x y)
+     #:with-this this
      (with-origin ctx [x y]
        (set-object! ctx
-                    [font          (js-string #js*.this.font)]
+                    [font          (js-string #js.this.font)]
                     [textAlign     #js"center"]
                     [textBaseline  #js"middle"]
-                    [fillStyle     (js-string #js*.this.color)])
-       (#js.ctx.fillText (js-string #js*.this.text) 0 0)))])
+                    [fillStyle     (js-string #js.this.color)])
+       (#js.ctx.fillText (js-string #js.this.text) 0 0)))])
 
 (define-proto Line
   (λ (x y pen-or-color)
-    (set-object! #js*.this
+    #:with-this this
+    (set-object! this
                  [type    "line"]
                  [width   (abs+ceil x)]
                  [height  (abs+ceil y)]
                  [mode    #f]
                  [pen     pen-or-color]))
   [render (λ (ctx x y)
+            #:with-this this
             (with-origin ctx [x y]
-              (with-path ctx {"outline" #js*.this.pen}
-                (define sx (- (abs (half #js*.this.x))))
-                (define sy (- (abs (half #js*.this.y))))
+              (with-path ctx {"outline" #js.this.pen}
+                (define sx (- (abs (half #js.this.x))))
+                (define sy (- (abs (half #js.this.y))))
                 (cond
                   [(and (>= x 0) (>= y 0))
                    (#js.ctx.moveTo sx sy)
@@ -202,7 +208,8 @@
 
 (define-proto Rectangle
   (λ (width height mode pen-or-color)
-    (set-object! #js*.this
+    #:with-this this
+    (set-object! this
       [type     "rectangle"]
       [width    width]
       [height   height]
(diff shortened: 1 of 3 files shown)


racketscript-extras/racketscript/private/jscommon.rkt
```diff
@@ -40,6 +41,14 @@
     #:description "a key-value pair for object"
     (pattern [name:id val:expr])))
 
+(define-syntax (field-λ stx)
+  (syntax-parse stx
+    [(_ formals (~datum #:with-this) self:id body ...)
```
glebm: Let's use an arrow function, then:
1. The rest of the code won't need to change.
2. `#:with-this` could be renamed to e.g. `#:bound` and won't need an argument.

vishesh: I would like the idea of using arrow functions. But there are two points to consider:

1. My understanding is that arrow functions are mostly convenience. Concise syntax and no `this` binding. Do we benefit same thing in Racket? 
  - Concise syntax can be achieved via reader extensions. 
  - The reason `this` is not bound for arrow functions is because its easy to make mistakes e.g. callbacks where `this` changes. By forcing programmer to explicitly bind the `this` pointer to some name in every case is easy to reason about unless (3).
2. Automatically, considering few kind of lambdas as arrows and others as otherwise is kind of confusion both for programmer and for compiler writer (particularly when doing optimizations). Specially with macros and all meta-programming stuff, it may lead to confusion on how this lambda will actually compile. So explicit annotation is important here IMO.
3. We can add a special form that compiles to arrow function. Do not entirely like the idea, as it requires adding a special FFI case again to say this lambda is an arrow function.

glebm: > My understanding is that arrow functions are mostly convenience.

Mostly, but not entirely. The arrow functions have simpler semantics than functions bound with `bind` (e.g. `bind` overrides `toString` on the function), that's why they are faster.

Here is a comprehensive benchmark: https://jsperf.com/arrow-function-vs-bind-vs-context-argument/4. Note especially the results for the current browser betas.

Additionally, they not only scope `this` lexically, but also other things like `super` too, which is simply not possible with `bind`.

> Automatically, considering few kind of lambdas as arrows and others as otherwise is kind of confusion both for programmer and for compiler writer.

Yeah, makes sense, nevermind then.

vishesh: So correct me if I'm wrong but ES6 classes are primarily a syntactic sugar right (at-least that's what I read last year)? In that case we don't have to worry about `super` for now. In our case we not using `bind`, but using a closure record which arrow syntax would do so internally as well.

> Automatically, considering few kind of lambdas as arrows and others

However, if these are faster than normal lambdas by good factor, a future possibility could be try proving where it is safe to use them and build optimizations based on that.

glebm: > So correct me if I'm wrong but ES6 classes are primarily a syntactic sugar right (at-least that's what I read last year)?

Mostly, but classes can be much faster, because:
1. `extend` and `super` are implemented natively.
2. A class can also only be declared once.

> In our case we not using bind, but using a closure record which arrow syntax would do so internally as well.

Yes, it doesn't actually help in this case (`const obj = { ... }`), I misunderstood the original issue this PR is trying to fix.

racketscript-extras/racketscript/htdp/image.rkt
```diff
@@ -519,31 +539,34 @@
...
      (with-origin ctx [x y]
-       (#js.ctx.rotate #js*.this.radians)
-       (#js*.this.image.render ctx 0 0)))])
+       (#js.ctx.rotate #js.this.radians)
+       (#js.this.image.render ctx 0 0)))])
 
 (define-proto Scale
   (λ (image x-factor y-factor)
-    (set-object! #js*.this
+    #:with-this this
```
glebm: This doesn't seem like it would work any different from before.

As I understand, this compiles to:

```js
Scale.prototype.image = function(...) {
  var self = this;
  // ...
};
```

The above doesn't actually bind the function:

```js
var fn = new Scale().image;
fn(1, 1); // `self` will be `window` here.
```

vishesh: That's a good point. I guess we carefully are using `Scale.image()` form everywhere. I actually did forgot about  `bind()` function when working on this patch, and when I meant bind I meant in more generic sense of binding variables not binding. For this patch it is ok, as like you said it is equivalent to before. The advantage is we are avoiding the unsafe `#js*.this` to some extent, and when callbacks inside this function are made, they will automatically have this/that way of doing things (as this will be bound to different name, and callback will hold it in its closure).

However, cases like `let fn = new Scale().image` will be harder to figure out and apply bind function automatically. For higher level object model like we have here, we can try some other way to initialize prototype chain so that correct this is available (e.g. use bind()).

glebm: > The advantage is we are avoiding the unsafe #js*.this to some extent,

In the case above, there is absolutely no difference at all in the behaviour vs the previous version, is there?

vishesh: Yeah. We can think it more as a Python way of OOP of handling this pointer.

Overall, main motive of this PR was to find and discuss exactly this kind of problem, that needs to be addressed in our FFI and programming model. I added an edit to last comment.

glebm: Alternatively we could just preserve JS semantics and let the user be in control of `this` (pun intended).

For example, one could use `(define self *this*)` instead of `#:with-this self`, and an explicit `#js(arg)=>` for arrow functions in the future.

This seems reasonable as JS semantics are not in conflict with Racket semantics here, as Racket does not have the concept of `this` at all.

glebm: > Just to add this semantics are really weird considering they persist even with classes

They are not weird if you think of it like this: There is no special handling for `this` and, other than the syntax sugar, it's just an argument like any other. All functions are free functions.

vishesh: > Alternatively, we could just preserve JS semantics and let the user be in control

Use actually can do that. Our FFI is layered, one layer is macro over another. What I would like to see is a model unique to RacketScript that provides more safety than that found in JavaScript. Otherwise, we are just a syntactic sugar over plain JavaScript, in which case one might as well write JavaScript. :)

> JavaScript is more Lisp than Java at heart.

Haha, I wish. I find it weird because not because I don't understand them, but their semantics simply don't justify their existence. I would like it either totally gone and objects passed explicitly (like C), or do it right and avoid all misery (like Java, Python ...). It is too low-level detail for a high level language.

> It's a common, if hackish, pattern in JavaScript to apply a method from one class to an instance of another.

This gives me shivers :P

glebm: Well for a higher-level model why not just use actual Racket classes, why a layer in-between?

vishesh: 1. Racket classes uses contracts (IIRC) and what not atm, and that will take
much longer than implementing a small class system ourselves.
(Contracts are important nevertheless).
2. Racket's implementation is likely to be very slow for us. Our
implementation can be specialized for JS target.

Whatever layer I was talking about, I was hoping it to be pretty much
like Racket ones. Perhaps we may be even be able to use our
implementation instead of Racket one for OO programs in Racket.

glebm: > Racket's implementation is likely to be very slow for us.

So I'm not very familiar with Racket classes, but I've had a look at the guide on the Racket's site and what I've seen seems to be implementable efficiently in JavaScript. Is there anything in particular you think would make them slow?

vishesh: Racket OOP is macros and expand to core forms. Try a OOP program and do `raco expand /path/to/oop/program.rkt`. You will see that in in the end its all just function application, lambdas, and the data structures we know (structs, vectors ...). We can implement Racket classes ourselves (which is probably huge task), or use existing implementation by supporting more Racket features (if you compile a OOP program first you will encounter #93). Latter is the one I was talking about and is likely to be slow for us (not using native DS, huge generated code ...).

Issue #143, opened 2018-09-14
RacketScript broken for Racket 7


vishesh:
Seems like some new modules and some things moved from `#%kernel` to `#%runtime`.

```
± % traceur js-build/modules/fact.rkt.js
Error: Error: ENOENT: no such file or directory, open '/home/vishesh/Racket/racketscript/tests/js-build/collects/racket/private/misc.rkt.js'
Specified as ../collects/racket/private/misc.rkt.js.
Imported by /home/vishesh/Racket/racketscript/tests/js-build/modules/fact.rkt.js.
Normalizes to /home/vishesh/Racket/racketscript/tests/js-build/collects/racket/private/misc.rkt.js
locate resolved against base '/home/vishesh/Racket/racketscript/tests/'

    at InternalLoader.handleCodeUnitLoadError (/home/vishesh/.nvm/versions/node/v10.10.0/lib/node_modules/traceur/bin/traceur.js:33360:40)
    at /home/vishesh/.nvm/versions/node/v10.10.0/lib/node_modules/traceur/bin/traceur.js:33222:20

```

brown131:
Is a fix forthcoming?

vishesh:
I'm not actively working in this fix at the moment. I think I will probably start work after couple of weeks. If you are interested I can share a WIP patch meanwhile.

Also, with a fix for this, I will drop support for 6.x. Racket 7 comes with bunch of changes including linklets, and wouldn't want to make things harder by supporting 6.x.

brown131:
Thanks for your answer! After trying to use GHCJS, I was afraid that this project was rusting as well. 

I'm all in favor of you fixing forward and focusing your energies on 7.0.

gregid:
Is there any update on the upgrade to Racket 7.0?

juszczakn:
I took a stab at this.

https://github.com/juszczakn/racketscript/commit/e200be0ea39aa058ee1d2db2b933eab196ed4059

It's kinda a hack, to get started. 

The next problem seems to bet that #%kernel isn't getting listed as a import for something. Not sure what or how that's happening. 

I was able to actually finish a compile on latest racket 7.2 by adding:
```
modified   racketscript-compiler/racketscript/compiler/expand.rkt
@@ -365,7 +365,7 @@
     #:literal-sets ((kernel-literals #:phase (current-phase)))
     [(module name:id lang:expr (#%plain-module-begin forms ...))
      (parameterize ([current-module path]
[...]

PR #148, opened 2019-09-11
fix test suite for Racket 7.x

stchang:
This pr fixes the tests for Racket 7.x (see https://travis-ci.org/stchang/racketscript/).

I'm not completely happy with it, but I figure I would submit this since there seems to some interest in just getting something working (see #143). I'll let @vishesh decide whether to wait or to merge this first.

These changes map the new primitive modules (eg `#%runtime`) to the existing `kernel.rkt`, and leaves empty placeholder files for the actual primitive modules.

Ideally, we would actually split kernel.rkt into the new modules but I'm a little stuck at the moment.


stchang:
A few more details about what I think is the current problem:
1. Building the initial dependency graph (`global-export-graph`) stops at `#%kernel` and never reaches `#%runtime`, etc.
2. But `#%runtime`, etc, is still reported as the source module by `identifier-binding`, so `#%kernel` never gets added to `module-object-name-map`.
3. This causes eventual failure in `ImportedIdent` case of `absyn-expr->il` in `transform.rkt` when looking up the `#%kernel` module.

PR #150, opened 2019-09-18
fix bug in string cmp fns (eg string<?); closes #149

### LINKED ISSUE #149: string cmp fns, eg string<? sometimes return wrong result
eg
`(string<? "b" "aa")` should be `#f` but returns `#t`

stchang:


racketscript-compiler/racketscript/compiler/runtime/core/unicode_string.js
```diff
@@ -434,16 +434,13 @@ export function eq(a, b) {
  * @return {!boolean}
  */
 export function lt(a, b) {
-    if (a.length < b.length) {
-        return true;
-    }
-    const n = a.length;
+    const n = (a.length < b.length) ? a.length : b.length;
```
glebm: Consider replacing with:

```js
Math.min(a.length, b.length)
```

It is more readable and also an order of magnitude faster in this microbenchmark on Firefox 69: https://jsperf.com/math-min-vs-if-condition-vs/13

stchang: Thanks, I'll fix it

Issue #106, opened 2017-09-06
Implement `weakhash*` primitives


vishesh:
As discussed in #105

stchang:
Running into some problems trying to implement `make-weak-hash` (the one that uses `equal?`).

Specifically, I can't figure out how to do a `hash-ref` when the given key is `equal?` but not `eq?` to one that is (or was) in the hash.

In this case, `hash-ref` should return the corresponding value if the key has not been collected yet but how can I figure out if the key has been collected?

Holding the keys in a `WeakMap` or `WeakSet` doesnt work since those are not iterable, and I cant use `has` since that compares with `eq?`.

(The fact that `WeakMaps` are not iterable also makes it hard to print a `weak-hash`, but that's a separate and more minor issue.)

Am I missing something obvious? Any suggestions?

vishesh:
I don't think you are missing anything. I'm convinced myself it is not possible with following reasoning.

`WeapMap` works on objects, and essentially what we want is to compute a unique identify for all values that are `equal?` at that point of execution. Hashing has two problems:

1. It generates a number for which we have to use a regular `Map`.
2. Even if `WeakMap` can take integers, we can't guarantee a unique hash for `equal?` values, regardless how many hashes you take (only make collisions less frequent). And like you said, to check for collisions we need iteration through another WeakMap collection mapped to the key, which we again can't do.

So, we need a mechanism to map our all `equal?` values to a unique identity object, which can be used when we do ref. Building such map is again equivalent to the original problem we have at hand. Overall, if a value is in WeakMap, one object of same value should be in a regular value based hash/set with which we have leak. We won't leak everything, i.e. out of 10 objects involved that are equal 1 would stay in memory forever, but it is still not correct.

stchang:
Ok I will implement `make-weak-hasheq` with `WeakMap`.

For `make-weak-hasheqv`, I think it could be a `WeakMap` but it depends on how we implement `eqv`? Specifically, are there number objects that are `eqv?` but not `eq?`.

For `make-weak-hash` is it better to error or default to non-weak hash (and leak memory)?

vishesh:
> For make-weak-hash is it better to error or default to non-weak hash (and leak memory)?

[...]

PR #151, opened 2019-09-24
add (part of) racket-core hash tests

stchang:


stchang:
Here are some more notes for this pr. Open to suggestions for any of the items below.

#### Remaining hash table features not implemented
- weak hashes (have not revisited since this discussion https://github.com/vishesh/racketscript/issues/106)
- fns in `racket/hash`, eg `hash-union`
- sort argument of `hash-map`
- chaperones and impersonators

#### Hash table features with different semantics in RacketScript
- iteration via `hash-iterate-first`/`next` 
  - indices are javascript objects, not numbers
  - deleting keys while iterating does not have the same semantics
    - mutating hash in RacketScript resets the iteration
- when overwriting a val for an existing key, when the keys are not `eq` (but equal according to the equality fn), Racket keeps the old key but RacketScript (ie hamt.js) uses the new key

#### Other
- `arithmetic-shift` does not work for numbers greater than 32 bit (I think)
- did not implement unreadable symbols

racketscript-compiler/racketscript/compiler/runtime/core/hash.js
```diff
@@ -84,20 +86,155 @@ class Hash extends PrintablePrimitive {
...
 
+    // implements hash-ref-key
+    refkey(k, fail) {
+        if (this._h.has(k)) {
+	    for (const key of this._h.keys()) {
+		if (hashConfigs[this._type].keyEq(key, k)) {
+		    return key;
+		}
+		// can't get here
+		return k;
```
vishesh: I don't think I understand this part. So `hash-ref-key` is supposed to return the key that is used internally (hence the `eq?`). Since, HAMT doesn't have any API to get that reference, we are iterating through all keys. But seems like we just return on first iteration of loop.

```javascript
if (this._h.has(k)) {
  for (const key of this._h.keys()) {
    if (hashConfigs[this._type].keyEq(key, k)) {
      return key;
    }
  }
  // can't get here
  return k;
}
```

If that is the case, I think its safer to have an `assert` or some runtime exception (`racketCoreError`?) instead of returning the original `k` where we say `// can't get here`.

racketscript-compiler/racketscript/compiler/runtime/core/hash.js
```diff
@@ -84,20 +86,155 @@ class Hash extends PrintablePrimitive {
...
+    isKeysSubset(v) {
+	if (!check(v)) {
+	    return false;
+	}
+
+        if (this._type !== v._type) {
+	    throw racketCoreError('hash-keys-subset?: ',
+				  'given hash tables do not use the same key comparison\n',
+				  'first table:', this);
+	}
```
vishesh: Unaligned indent. There are some more in file it seems. There seems to be `eslint` file here https://github.com/vishesh/racketscript/blob/master/.eslintrc, but I kind of forgot how I used the tool.

racketscript-compiler/racketscript/compiler/runtime/core/hash.js
```diff
@@ -84,20 +86,155 @@ class Hash extends PrintablePrimitive {
...
+    iteratePair(i) {
+        return Pair.make(i.value[0], i.value[1]);
+    }
+
+    iterateKeyValue(i) {
+        return Values.make(i.value);
+    }
+
+    union(h) {
+        let newH = this._h;
```
vishesh: Unaligned indent.

racketscript-compiler/racketscript/compiler/runtime/core/hash.js
```diff
@@ -84,20 +86,155 @@ class Hash extends PrintablePrimitive {
...
     size() {
         return this._h.size;
     }
 
+    // iteration operations, eg hash-iterate-first/next
+    iterateFirst() {
+	if (this._h.size == 0) return false;
+	// must save iterator since next() is stateful
+        this._iterator = this._h.entries();
+	return this._iterator.next();
```
vishesh: We don't support mutation while iterating for now, but I was wondering if there any value for supporting multiple iterators?

Also PR notes mention that mutating keys resets the iterator. If we pull out the iterator in its own class, and hold a reference to the hash table it is iterating, can we retain the original Racket behaviour? If my assumption is right, for an immutable hash, the HAMT iterator should work as it would in Racket. For mutable I think Racket should consider the next added/removed keys, and in case of RacketScript we can maybe check if the HAMT reference in underlying reference has changed and invalidate it?

We can skip however ignore this for now, and just take note for future. Overall, seems like HAMT adds quite a few hurdles to do Racket hashes efficiently/correctly?

racketscript-compiler/racketscript/compiler/runtime/core/hash.js
vishesh: Ah ok, I think your PR notes explain why for this -- 

    when overwriting a val for an existing key, when the keys are not eq (but equal according to the equality fn), Racket keeps the old key but RacketScript (ie hamt.js) uses the new key

Which is what seems to be issue here?

stchang: Oops, I think there is a misplaced closing brace. The `return k` should be outside the `for` loop. Does that clear things up?

racketscript-compiler/racketscript/compiler/runtime/core/hash.js
stchang: > if there any value for supporting multiple iterators

Yes, that would be ideal

> If we pull out the iterator in its own class, and hold a reference to the hash table it is iterating, can we retain the original Racket behaviour?

It might be closer to but it still wont match Racket's semantics, since the Racket iterator is essentially a pointer to the memory of the current element.

> Overall, seems like HAMT adds quite a few hurdles to do Racket hashes efficiently/correctly?

Most of the workarounds are for mutable hashes. Can we use a different internal representation for mutable hashes, like Racket does? Could we just use a plain Javascript map?

racketscript-compiler/racketscript/compiler/runtime/core/hash.js
stchang: Yes I was having some trouble with the indentation. For some reason the default in emacs is inserting tabs instead of spaces. Do you use any specific settings? Or a formatter?

racketscript-compiler/racketscript/compiler/runtime/core/hash.js
stchang: Ah ok thanks, I'll try to figure out how to incorporate those settings

racketscript-compiler/racketscript/compiler/runtime/core/hash.js
vishesh: So a quick look at my config shows that I used `js2-mode`, but no specific hooks to override tabs/space, but I do have my emacs configured to universally expand tabs regardless of language. If my memory serves right, there is some integration for eslint.

racketscript-compiler/racketscript/compiler/runtime/core/hash.js
vishesh: I think we'll have to use some sort of library (or spin up our own) that takes equality operator like HAMT (like some RB-tree implementation). The plain JavaScript map just takes strings as keys, and the `Map` class uses '===' operator.

vishesh:
Apologies for slow review cycle here. I hope they are not blocking you. So overall, the PR looks good to me, and once the brace placement issue is fixed we can merge.  :)

Also, if you have some notes already on how you are going after Racket 7, it would be great if you can do a quick dump on the  Racket 7 issue? That would potentially reduce some duplicate effort by me to figure the action items by me again.

stchang:
Take your time. I appreciate whatever effort you can spare.

I'll try to log more notes so we can stay on the same page.

racketscript-compiler/racketscript/compiler/runtime/core/hash.js
stchang: Ok I see. (I was looking at the wrong hamt.js docs.) For some reason I thought the equal fn was only part of our `Hash` class but I see now that `hamt.js` uses it directly.

Issue #153, opened 2019-10-02
procedure variable arity?


stchang:
I'm currently trying to improve error handling to better match racket and that involves implementing `procedure-arity-includes?` which gets called a lot.

One thing I ran into is that I think it's currently not possible to identify variable arity functions. Is that right?

Can we do this by adding another field, similar to [`__rjs_arityValue` here](https://github.com/vishesh/racketscript/blob/master/racketscript-compiler/racketscript/compiler/runtime/core/procedure.js)?

vishesh:
It works IMO right now, but there seems to be a bug.

```racket
#lang racket

(equal? (procedure-arity (lambda (x) x)) 1)
;;> #t

(displayln (procedure-arity (lambda (x . y) x)))
;;> #(struct:kernel:arity-at-least 1)

(equal? (procedure-arity (lambda (x . y) x) (arity-at-least 1)))
;;> #f  -- WRONG: should be #t
```
The `arity-at-least` is computed correctly. IIRC looking at definition of[ `procedure-arity`](https://github.com/vishesh/racketscript/blob/master/racketscript-compiler/racketscript/compiler/runtime/kernel.rkt#L1033) the valu of `__rjs_arityValue` is supposed to be an array when its variable.

One problem however I see is in the last case of above example. Somehow, `arity-at-least` is not equal to `kernel:arity-at-least`.

stchang:
oh i see. what about primitives like `values` or `list`? calling `procedure-arity` returns the result of the `length` field, which is 0.

would calling `attachProcedureArity` around the definition fix this?

vishesh:
Yes, I think any `lambda` with variable arity should be wrapped with `attachProcedureArity`. Missed doing that for kernel primitives.

Currently almost all kernel primitive use `define+provide` so should be an easy fix. We can perhaps introduce `define-kernel-primitive` or something so that any internal define doesn't have that overhead?

PR #156, opened 2019-10-07
Exceptions

### LINKED ISSUE #152: unifying implementation of exceptions
Currently, I think `error` compiles to `throw`, but from what I can tell, nothing compiles to `catch`. I think this means that I can't write tests that expect an error?

Separately, there is the (still experimental?) `raise`, which can be caught using `with-handlers`, both of which compile to an implementation of continuation marks.

Was the plan to have the second continuation-marks-based implementation be the eventual default case?

While implementing tests, I've so far skipped writing tests that expect an error but it would be good to enable them at some point.

stchang:
### This pr:
- makes all exns catchable via `with-handlers`
  - requires moving most error throwing to `kernel.rkt` (is this the right thing to do?)
- cleans up error api to match racket's api, including exact error message formatting
  - use this api, eg `raise-argument-error`, etc, in more places
- allows writing tests that expect an exn
  - enable these kinds of tests in `racket-core` 

#### smaller changes
- manually attach arity and name to prim fns (is this right?)
- allows calling cmp fns (eg `<`) with one arg (added in Racket7 I think)

#### other notes
- closes #152 and #153 
- needs squashing before merge (I can do it if that's easier)

### DIFF
--- a/racketscript-compiler/racketscript/compiler/assembler.rkt
+++ b/racketscript-compiler/racketscript/compiler/assembler.rkt
@@ -74,6 +74,7 @@
      (emit-args args ",")
      (emit ")")]
     [(ILBinaryOp oper args)
+     (assert (>= (length args) 2)) ;; TODO: Update the type of ILBinaryOp
      (for/last? ([arg last? args])
        (when (ILBinaryOp? arg) (emit "("))
        (assemble-expr arg out)
--- a/racketscript-compiler/racketscript/compiler/runtime/kernel.rkt
+++ b/racketscript-compiler/racketscript/compiler/runtime/kernel.rkt
@@ -90,9 +90,9 @@
 
 (define+provide (single-flonum-available?) #f)
 
-(define+provide *  (#js.Core.attachProcedureArity #js.Core.Number.mul 1))
+(define+provide *  (#js.Core.attachProcedureArity #js.Core.Number.mul 0))
 (define+provide /  (#js.Core.attachProcedureArity #js.Core.Number.div 1))
-(define+provide +  (#js.Core.attachProcedureArity #js.Core.Number.add 1))
+(define+provide +  (#js.Core.attachProcedureArity #js.Core.Number.add 0))
 (define+provide -  (#js.Core.attachProcedureArity #js.Core.Number.sub 1))
 (define+provide <  (#js.Core.attachProcedureArity #js.Core.Number.lt 1))
 (define+provide >  (#js.Core.attachProcedureArity #js.Core.Number.gt 1))
--- a/racketscript-compiler/racketscript/compiler/runtime/kernel.rkt
+++ b/racketscript-compiler/racketscript/compiler/runtime/kernel.rkt
@@ -170,8 +170,10 @@
 
 (define-checked+provide (car [pair pair?]) #js.pair.hd)
 (define-checked+provide (cdr [pair pair?]) #js.pair.tl)
-(define+provide cons (#js.Core.attachProcedureName #js.Core.Pair.make "cons"))
-(define+provide pair?      #js.Core.Pair.check)
+(define+provide cons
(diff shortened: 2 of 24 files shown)


racketscript-compiler/racketscript/compiler/runtime/kernel.rkt
```diff
@@ -88,15 +90,15 @@
...
-(define+provide +  #js.Core.Number.add)
-(define+provide -  #js.Core.Number.sub)
-(define+provide <  #js.Core.Number.lt)
-(define+provide >  #js.Core.Number.gt)
-(define+provide <= #js.Core.Number.lte)
-(define+provide >= #js.Core.Number.gte)
-(define+provide =  #js.Core.Number.equals)
+(define+provide *  (#js.Core.attachProcedureArity #js.Core.Number.mul 1))
+(define+provide /  (#js.Core.attachProcedureArity #js.Core.Number.div 1))
+(define+provide +  (#js.Core.attachProcedureArity #js.Core.Number.add 1))
```
vishesh: `+` and `*` in Racket have arity of `(arity-at-least 0)`.

I'm thinking a good way to keep Racket and RacketScript primitives in sync would be to define a macro `define-primitive` which can during macro expansion phase look at the primitive arity and check it against what we give (or something similar). That way we can maybe also track a bit more easily which primitives are not exactly like Racket as we wont be able to use `define-primitive` for that?

stchang: I did try arity 0, but got a `case-lambda: invalid case` error. Do you know if that means `case-lambda` doesnt allow 0 args, or I just need to change the definition of `+`?

stchang: Also, good idea with `define-primitive`

vishesh: Its a compiler bug actually, and case-lambda errors shows up because of displayln ends up getting no expression due to the bug. I've created a PR against your branch.

stchang: ah nice thanks

racketscript-compiler/racketscript/compiler/assembler.rkt
```diff
@@ -74,6 +74,7 @@
      (emit-args args ",")
      (emit ")")]
     [(ILBinaryOp oper args)
+     (assert (>= (length args) 2)) ;; TODO: Update the type of ILBinaryOp
```
stchang: hmm, the tests fail here (but only for Racket <= 6.8). investigating

stchang: Could the problem [be these lines](https://github.com/vishesh/racketscript/blob/master/racketscript-compiler/racketscript/compiler/transform.rkt#L422-L424)?

stchang: Actually, all versions were failing. [I pushed something that fixes the tests](https://github.com/vishesh/racketscript/pull/156/commits/b7ae80a9c87a1c90bed3167b9771ea5e95245586) but let me know if you think that's the wrong place

stchang: actually, should [the last commit](https://github.com/vishesh/racketscript/pull/156/commits/b7ae80a9c87a1c90bed3167b9771ea5e95245586) use `>=` instead of `=`, to match the assertion?

stchang: ok fixed

vishesh: Thanks! You are right.

stchang:
Didn't add `define-primitive` yet, but will do that separately :)

PR #158, opened 2019-10-10
Fix NPM security audit warnings

vishesh:
Updates Gulp version and top-level `package.lock`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,7 +14,7 @@
       "integrity": "sha1-r9+UiPsezvyDSPb7IvRk4ypYs2s=",
       "dev": true,
       "requires": {
-        "acorn": "3.3.0"
+        "acorn": "^3.0.4"
       },
       "dependencies": {
         "acorn": {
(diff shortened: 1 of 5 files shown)


package-lock.json
```diff
@@ -586,12 +635,12 @@
...
       "requires": {
-        "fs.realpath": "1.0.0",
-        "inflight": "1.0.6",
-        "inherits": "2.0.3",
-        "minimatch": "3.0.4",
-        "once": "1.4.0",
-        "path-is-absolute": "1.0.1"
+        "fs.realpath": "^1.0.0",
+        "inflight": "^1.0.4",
+        "inherits": "2",
```
stchang: Should this have caret `^`?

package-lock.json
```diff
@@ -674,8 +723,8 @@
       "integrity": "sha1-Sb1jMdfQLQwJvJEKEHW6gWW1bfk=",
       "dev": true,
       "requires": {
-        "once": "1.4.0",
-        "wrappy": "1.0.2"
+        "once": "^1.3.0",
+        "wrappy": "1"
```
stchang: same question as above

package-lock.json
```diff
@@ -955,7 +996,7 @@
       "integrity": "sha1-WDsap3WWHUsROsF9nFC6753Xa9E=",
       "dev": true,
       "requires": {
-        "wrappy": "1.0.2"
+        "wrappy": "1"
```
stchang: same question as above

stchang:
Dont have much experience with these config files but I reinstalled and tests seem to be passing.

stchang:
Just had one question about maybe missing carets?

stchang:
ah ok looks good then

vishesh:
Oups, misread your question. You asked if we should add caret. I think overall it should be fine, as same points apply.

PR #168, opened 2020-04-15
build: Fixes babel build

vishesh:
Issue was outdated gulp/babel build files.


vishesh:
Seems like 6.4 and 6.8 are already failing in master.

stchang:
yes, something changed with the `cover` pkg that made racket 6.x unhappy

PR #171, opened 2020-08-23
Implement more char functions

### LINKED ISSUE #170: Implement Unicode property testing functions (e.g. `char-alphabetic?`)
Now that browsers support [Unicode property escapes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Unicode_Property_Escapes), we can finally implement Unicode property testing functions in RacketScript.

glebm:
1. All `char-*?` Unicode property test methods.
2. `char-upcase` and `char-downcase`.

Fixes #170


vishesh:
Thanks for the patch! Looks good to me.

PR #182, opened 2021-02-04
Allow all identifiers in VarRef.

samth:


stchang:
Thanks. I tried this too but the compiled js file still fails. We can discuss in a few min.

```
(function (exports, require, module, __filename, __dirname) { import * as $rjs_core from '../runtime/core                .js';import * as M0 from "../collects/racket/private/for.rkt.js";import * as M1 from "../collects/racket/                private/map.rkt.js";import * as M2 from "../runtime/paramz.rkt.js";import * as M3 from "./testing.rkt.js"                ;import * as M4 from "../collects/racket/private/misc.rkt.js";import * as M5 from "../collects/racket/pri                vate/modbeg.rkt.js";import * as M6 from "../collects/racket/private/more-scheme.rkt.js";import * as M7 fr                om "../collects/racket/private/hash.rkt.js";import * as M8 from "../runtime/kernel.rkt.js";import * as M9                 from "../collects/racket/private/list.rkt.js";import * as M10 from "../collects/racket/private/reverse.r                kt.js";import * as M11 from "../runtime/unsafe.rkt.js";var val__gt_type_num = function(v1) {if (arguments                .length!==1) {throw $rjs_core.racketContractError("arity mismatch");} else {}if (M8.boolean_p(v1)!==false                ) {var if_res11 = 0

SyntaxError: Unexpected token import
    at createScript (vm.js:80:10)
    at Object.runInThisContext (vm.js:139:10)
    at Module._compile (module.js:616:28)
    at Object.Module._extensions..js (module.js:663:10)
    at Module.load (module.js:565:32)
    at tryModuleLoad (module.js:505:12)
    at Function.Module._load (module.js:497:3)
    at Function.Module.runMain (module.js:693:10)
    at startup (bootstrap_node.js:188:16)
    at bootstrap_node.js:609:3
```

PR #191, opened 2021-05-18
Add GH Actions for running eslint, unit and integration tests

daynin:


stchang:
Thanks for starting to set these up. Let us know if we can help, or answer any questions. I'm not fluent with gh actions yet, but does the integration test workflow need to install js?

daynin:
@stchang thank you! But I've finished just now

> I'm not fluent with gh actions yet, but does the integration test workflow need to install js?

No, there will be two separate actions (for elsint and for racket tests) and they will install their deps independently.

Btw, looks like tests pass almost for every racket version (excluding 6.9 and 8.1).

racketscript-compiler/info.rkt
```diff
@@ -19,6 +19,7 @@
 
 (define pkg-authors '(vishesh))
 (define pkg-desc "Racket to JavaScript compiler")
+(define post-install-collection "")
```
daynin, line 22: I don't know why but without that variable installation failed in GH Actions

stchang:
Ah great. Let me know when this is ready to merge

racketscript-compiler/info.rkt
stchang: Weird, I havent seen this before. Thanks for figuring it out.

daynin:
@stchang looks like, now it's ready

daynin:
But pay an attention please, I deleted the TravisCI config

stchang:
> But pay an attention please, I deleted the TravisCI config

Yes, that's good. We wanted to get rid of it anyways

vishesh:
This is awesome! Thanks!!

Issue #189, opened 2021-05-17
racket 8.0 regressions


stchang:
[ ] `and` vs `and/c` contract?
[ ] some difference in hash fn args

diff:
```
$ diff <(racket tests/racket-core/hash.rkt) <(node js-build/modules/hash.rkt.js)
123c123
<   expected: (and/c hash? immutable?)
---
>   expected: (and hash? immutable?)
124a125,128
>   argument position: 1st
>   other arguments...:
>    1
>    2
127a132,134
>   argument position: 1st
>   other arguments...:
>    1
130a138,141
>   argument position: 1st
>   other arguments...:
>    1
>    2
133a145,147
>   argument position: 1st
>   other arguments...:
>    1
136c150
<   first table: '#hasheqv()
---
>   second table: '#hasheqv()
139c153
<   first table: '#hasheq()
---
>   second table: '#hasheq()
142c156
<   first table: '#hasheq()
---
>   second table: '#hasheq()
145c159
<   first table: '#hasheqv()
---
>   second table: '#hasheqv()
```

vishesh:
@stchang : https://github.com/vishesh/racketscript/pull/197 should fix this as well?

stchang:
yes, it should have fixed it, but in reality it did not. I just submitted another pr

vishesh:
Fixed in https://github.com/vishesh/racketscript/pull/199

PR #194, opened 2021-05-18
add timeout-minutes to avoid endless tests on Racket 8.*

daynin:
- add timeout
- delete racket 6.9
- add racket 8.0

### DIFF
--- a/.github/workflows/racket.yml
+++ b/.github/workflows/racket.yml
@@ -9,11 +9,18 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
+    continue-on-error: ${{ matrix.experimental }}
     timeout-minutes: 20
     strategy:
       fail-fast: false 
       matrix:
-        racket-version: [ '6.12', '7.0', '7.4', '7.5', '7.6', '7.7', '7.8', '8.0', 'stable' ]
+        racket-version: [ '6.12', '7.0', '7.4', '7.5', '7.6', '7.7', '7.8' ]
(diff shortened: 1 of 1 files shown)


stchang:
Did you mention that tests are passing with Racket 8.0 now? Should 8.0 be added to the matrix?

stchang:
Actually some tests still fail for me using 8.0. I started a list here: https://github.com/vishesh/racketscript/issues/189

daynin:
No, I didn't run tests on Racket 8.0. Added 8.0 version to the matrix

stchang:
Do you know if there a way to "allow failures" in the gh actions config like you could with travis?

daynin:
@stchang it can be a little bit tricky because GH Actions don't have the same feature, but I'll try to do that. You want to add an ability to skip fails of tests which run on 8.* version of Racket, right (just to be on the same page)?

stchang:
Yes, allow failure of 8.0 and 8.1 right now, until we can fix it. Although it looks like the workflow gets marked as passing right now, even if some tests fail: https://github.com/vishesh/racketscript/issues/195

daynin:
Ok, I'll try to do that

daynin:
Now it has success status even if 8.* versions have failed tests or other errors https://github.com/vishesh/racketscript/actions/runs/853924052. But I don't know why here in PR it looks like failed

vishesh:
Seems like 8.0 failures can be addressed separately, so I will just merge this.

Issue #195, opened 2021-05-18
make failing tests fail the workflow run


stchang:
Right now, a failing test does not make the GH action fail, e.g., https://github.com/vishesh/racketscript/pull/194/checks?check_run_id=2611674889

We might need to modify the test harness to do this.

daynin:
I'll try to add returning a correct exit code on fail

stchang:
fixed with #196

PR #196, opened 2021-05-18
return a correct exit code on tests fail

daynin:
It should help to avoid issues like that https://github.com/vishesh/racketscript/issues/195

### DIFF
--- a/tests/fixture.rkt
+++ b/tests/fixture.rkt
@@ -163,9 +163,10 @@
   (when (clean-output-before-test)
     (delete-directory/files (output-directory))))
 
-;; (Listof Glob-Pattern) -> Void
+;; (Listof Glob-Pattern) -> Boolean
 ;; If tc-search-patterns is simply a path to directory, run all test
-;; cases in that directory otherwise use glob pattern
+;; cases in that directory otherwise use glob pattern. 
+;; Returns #t or #f depending on test results
 (define (run-tests tc-search-patterns)
   ;; First clean the compiled modules always, to avoid cases where
   ;; compilation fails but it anyway proceeds with last module output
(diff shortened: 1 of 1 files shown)


tests/fixture.rkt
```diff
@@ -277,7 +284,9 @@
     (displayln "--------------------------------")
     (displayln "::: Optimizations on ::: All :::")
     (displayln "-------------------------------")
-    (run-tests tc-search-patterns)))
+    (set-passed! (run-tests tc-search-patterns)))
+
+  (when (not passed) (exit 1)))
```
vishesh, line 289:
```suggestion
  (unless passed (exit 1)))
```

tests/fixture.rkt
```diff
@@ -253,9 +260,9 @@
       (displayln "---------------------------------")
       (displayln "::: Optimizations on ::: none :::")
       (displayln "---------------------------------")
-      (run-tests (filter-not
+      (set-passed! (run-tests (filter-not
```
vishesh, line 263: I think this s-exp is not indented after change.

daynin: fixed

vishesh:
Hi @daynin,

Given recent fixes on master, can you rebase this PR against master, so that CI passes before we merge it. Thanks!

daynin:
@vishesh done

daynin:
I think we already can merge with PR because it's only about fixing the testing tool, so it'll help with future work on those errors. @vishesh what do you think?

vishesh:
Oh! My bad. It is fixing the fact failures are showing up in CI report. I mixed up which PR is doing what.

PR #202, opened 2021-05-20
Fix eslint warnings and refactor hamt.js (a little bit)

daynin:


racketscript-compiler/racketscript/compiler/runtime/core/hamt.js
```diff
@@ -721,65 +712,63 @@ Map.prototype.set = function (key, value) {
...
-};
+hamt.removeHash = (hash, key, map) => hamt.modifyHash(del, hash, key, map);
 
-Map.prototype.removeHash = Map.prototype.deleteHash = function (hash, key) {
-    return removeHash(hash, key, this);
+Map.prototype.removeHash = function (hash, key) {
+    return hamt.removeHash(hash, key, this);
 };
 
+Map.prototype.deleteHash = Map.prototype.removeHash;
```
daynin, line 721: Looks like we don't need `deleteHash` and `delete` (see below). At least all tests pass after deleting it. Is it for backward compatibility?

vishesh:
I wouldn't worry too much about HAMT.js, as its a third-party library without any changes made by us (If I recall correctly). It is I think unmaintained, but keeping it same will make porting new changes here easier..

Issue #193, opened 2021-05-18
Add versioning system


daynin:
I think we should create changelogs and follow some versioning system. It'll allow us to change supported versions of Racket and Node.js for every single major version of RacketScript.

stchang:
Open to suggestions, but otherwise we'll just follow Racket's major-minor-patch versioning system.

daynin:
I'm not sure we can completely rely on racket versions because we have node.js as a second major tool. Thus, maybe it's better to have our versioning system just to have an ability to change versions when we need to do that. Because future users of RacketScript will rely on the RacketScript's versions mostly.

stchang:
Oh I didn't mean to match Racket's exact version. Just to use Racket's version numbering conventions.

daynin:
Oh, I see. Yeah I think it's a great idea!

PR #207, opened 2021-06-03
feat: add changelog auto-generation

daynin:
Here a GH action config for changelog generation. It'll help to create release notes for every pushed tag (here you can see how it works https://github.com/github-tools/github-release-notes/releases). 

And by changing a source of the racketscript package here https://pkgs.racket-lang.org/package/racketscript from this pattern `https://github.com/‹user›/‹package›.git` to this one `https://github.com/‹user›/‹package›.git#tag` (see "GitHub Deployment" here https://docs.racket-lang.org/pkg/getting-started.html#%28part._github-deploy%29) we'll allow to install a particular version of racketscript.

Resolves: https://github.com/vishesh/racketscript/issues/193

But to make it work @vishesh should create CHANGELOG_TOKEN (see https://github.com/github-tools/github-release-notes#setup) and change the source path of the package.

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -1,0 +1,25 @@
+name: changelog
+
+on:
+  push:
+    tags:
+      - 'v*'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout branch
+      uses: actions/checkout@v1
+    - name: Use Node.js 10
+      uses: actions/setup-node@v1
+    - name: Install github-release-notes and generate changelog
+      run: |
+        npx github-release-notes@0.17.1 changelog --generate --override --token=${{ secrets.CHANGELOG_TOKEN }} --changelog-filename=TEMP_CHANGELOG.md
+      env:
+        CI: true
+    - name: Create and publish realese
+      uses: ncipollo/release-action@v1
+      with:
+        bodyFile: 'TEMP_CHANGELOG.md'
+        token: ${{ secrets.CHANGELOG_TOKEN }}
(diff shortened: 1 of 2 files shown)


vishesh:
Umm, sorry about not looking at it sooner. Looks good to me. What I don't understand is why it requires token. How will Github workflow get the token? Obv we can't make it part of the repository itself, as its supposed to be private (the token gives access to all private repos as well).

vishesh:
It appears there is `GITHUB_TOKEN` which is automatically generated and can be used here instead, and if I understand correctly, it just has current repository in its scope instead of wider scoped private token I can create. However, I'm still little anxious about PRs/CI being able to change things (from security pov) which you don't see in the diff itself.

daynin:
Oh, you're right. But let separate two different issues:
1. The first one is about potential access to private repos. This issue can be solved by creating an organization in Github for RacketScript. Since we'll have in that organization only public repos we won't need to care about private repositories and (looks like) we'll be able to better control keys (because in organization settings it's possible to specify repositories for tokens). And moreover, it'll be useful for some additional repositories in the future like a website repository for RacketScript or some additional tools.
2. The second issue is about trust in Github Release Notes actions. So here we can only analyze the source code of a particular version of the project and check that it doesn't do anything else except generating release notes.

If we aren't ready or don't want to do that right now, we can get rid of that github action and write changelogs manually (or don't write them at all) and use tags for releases.

I'm not sure which option is better. @vishesh what do you think?

vishesh:
I'm just gonna let this PR sit around for some time, until we can get to thinking about that. Among the points you mentioned, another problem is that none of us are following any specific commit guidelines (log message conventions, squashing commits, when to do releases etc...). I totally would love to have all that, along with proper processes, but once we have some pace and features/fixes in.

But if everyone else feels strongly about having this sooner, we can get this in.

daynin:
I think we can add this automation later when we'll be ready. It's not a problem at all

PR #209, opened 2021-06-09
paramz.rkt typo

stchang:


vishesh (approved):

vishesh:
Seems like need more tests for exceptions.

stchang:
If we include the playground examples in the tests, it will catch a bunch of things I think.

PR #221, opened 2021-06-15
update nothing.rkt stub; fixes stdin compile for playground

stchang:
Not sure if this is the right way to fix things, but compiling in the playground doesnt work for me without this change.

### DIFF
--- a/racketscript-compiler/racketscript/compiler/expand.rkt
+++ b/racketscript-compiler/racketscript/compiler/expand.rkt
@@ -310,7 +310,7 @@
                                             nom-src-mod-path-orig
                                             mod-src-id))]))
 
-           ;; If the moduele is renamed use the id name used at the importing
+           ;; If the module is renamed use the id name used at the importing
            ;; module rather than defining module. Since renamed, module currently
            ;; are #%kernel which we write ourselves in JS we prefer original name.
            ;; TODO: We potentially might have clashes, but its unlikely.
@@ -329,7 +329,6 @@
                [else
                 (match-let ([(cons (app last mod) (? symbol? id)) path-to-symbol])
                   (values id mod #t))]))
-
            (ImportedIdent effective-id effective-mod reachable?)])])]
     [(define-syntaxes (i ...) b) #f]
     [(set! s e)
--- a/racketscript-compiler/racketscript/compiler/nothing.rkt
+++ b/racketscript-compiler/racketscript/compiler/nothing.rkt
@@ -1,3 +1,6 @@
-#lang racket
+#lang racketscript/base
 
-(require "../interop.rkt")
+(require racketscript/htdp/image
+         racketscript/htdp/universe
+         racketscript/browser
+         racketscript/interop)


stchang:
To give some more details, when compiling with `--stdin`, some ids get marked as unreachable.

For example, without this change, when compiling [the blank example](https://github.com/vishesh/racketscript-playground/blob/master/examples/blank.rkt) from the playground, compiling the file directly produces:
```
$ racks --js --js-beautify examples/blank.rkt
import * as $rjs_core from '../runtime/core.js';
import * as M0 from "../collects/racket/private/misc.rkt.js";
M0.displayln($rjs_core.UString.make("Hello World!"));
var __rjs_quoted__ = {};
export {
  __rjs_quoted__
};
```

while piping the same file to stdin cant find `displayln` and produces
```
$ cat examples/blank.rkt|racks --js --js-beautify --stdin
import * as $rjs_core from '../runtime/core.js';
import * as M0 from "../collects/racket/private/misc.rkt.js";
M0.__rjs_quoted__.displayln($rjs_core.UString.make("Hello World!"));
var __rjs_quoted__ = {};
export {
  __rjs_quoted__
};
```

stchang:
Not sure whether this change is better, or whether [the stub file from the playground](https://github.com/vishesh/racketscript-playground/blob/master/stub.rkt) needs updating

vishesh:
I think the build will fail because of big-bang stuff being in separate package. I think what we can do for now is create "nothing.rkt" during runtime in tmp directory.

If there a Racket function to just resolve a module from s-expression, and get list of imports, I think building the graph should be fairly straightforward after that. Currently, `get-module-deps` is basically a tree traversal that starts from the input module, and instead we can just start from list of imports if we can get those.

stchang:
> I think the build will fail because of big-bang stuff being in separate package. I think what we can do for now is create "nothing.rkt" during runtime in tmp directory.
> 
> If there a Racket function to just resolve a module from s-expression, and get list of imports, I think building the graph should be fairly straightforward after that. Currently, `get-module-deps` is basically a tree traversal that starts from the input module, and instead we can just start from list of imports if we can get those.

oh yeah, I see that. 

Instead of having the `nothing.rkt` file, can we just directly call `get-export-tree` with something like `racket/base`?

vishesh:
Yes, we can do that, but we'll miss the big-bang libraries. So it may be better if can do something like:

```
(module-compile-imports (compile-module (expand (read-syntax-from-stdin))))
```

I just did quick experiment and seems like this will work out. I can make a fix later evening after work, or if have time you can take this over. 

```racket
       (read-accept-reader #t)
       (read-accept-lang #t)
       
       ;; Using compile-module-imports
       (parameterize ([current-namespace (make-base-namespace)])
         (define input2 (compile (read (current-input-port))))
         (pretty-print (module-compiled-imports input2)))
         
       ;; Using module->imports
       (parameterize ([current-namespace (make-base-namespace)])
         (define input (read (current-input-port)))
         (pretty-print input)
         (eval input)
         (pretty-print (module->imports ''anonymous-module)))
```


Diff for just quickly checking

```diff
diff --git a/racketscript-compiler/racketscript/compiler/main.rkt b/racketscript-compiler/racketscript/compiler/main.rkt
index f1db1f9..2816e7c 100644
--- a/racketscript-compiler/racketscript/compiler/main.rkt
+++ b/racketscript-compiler/racketscript/compiler/main.rkt
@@ -384,9 +384,24 @@
        ;; need stdin from playground, its fine for now.
        ;; TODO: Figure out a way to compile this syntax to
        ;; module code bytecode
+       (read-accept-reader #t)
+       (read-accept-lang #t)
+       #;(parameterize ([current-namespace (make-base-namespace)])
+         (define input2 (compile (read (current-input-port))))
+         (pretty-print (module-compiled-imports input2)))
+       (parameterize ([current-namespace (make-base-namespace)])
+         (define input (read (current-input-port)))
+         (pretty-print input)
+         (eval input)
+         (pretty-print (module->imports ''anonymous-module)))
+       (define mod-string (port->string (current-input-port)))
+       (read-accept-reader #t)
+       (read-accept-lang #t)
+       (define input (open-input-string mod-string))
+       (pretty-print  (expand (read-syntax input)))
        (global-export-graph (get-export-tree (build-path racketscript-compiler-dir
                                                          "nothing.rkt")))
-       (read-and-expand-module (current-input-port))]
+       (read-and-expand-module (open-input-string mod-string))]
       [else
        (quick-expand source)]))
```

stchang:
superseded by https://github.com/vishesh/racketscript/pull/222

PR #222, opened 2021-06-15
Prepare proper module dependency graph when taking module from stdin

### LINKED ISSUE #134: Compiling JS from stdin doesn't prepare id graph based on the imported modules
Original ticket: https://github.com/vishesh/racketscript-playground/issues/12

Currently, we just pick imported modules from `nothing.rkt`. Temporary workaround is to add some requires to `racketscript/compiler/nothing.rkt ` so that identifier graph includes these commonly used modules:

```racket
#lang racket

(require racket
         racketscript/htdp/image
         racketscript/htdp/universe
         (except-in racketscript/base #%module-begin)
         "../interop.rkt")
```

Actual fix would require changing `get-export-tree` in `moddeps.rkt` to be able to take compiled code or module path. `get-module-deps` figures out what modules are imported by using (racket) compiled module, and for stdin input we can compile the syntax itself to bytecode and use same `compiled-module-imports` on that.

vishesh:
Currently, we relied on a stub "nothing.rkt", but we should instead look at the
imported modules itself. This should fix the the 'undefined identifier' issues
we keep getting in Playground.

FIXES #134

### DIFF
--- a/racketscript-compiler/racketscript/compiler/expand.rkt
+++ b/racketscript-compiler/racketscript/compiler/expand.rkt
@@ -310,7 +310,7 @@
                                             nom-src-mod-path-orig
                                             mod-src-id))]))
 
-           ;; If the moduele is renamed use the id name used at the importing
+           ;; If the module is renamed use the id name used at the importing
            ;; module rather than defining module. Since renamed, module currently
            ;; are #%kernel which we write ourselves in JS we prefer original name.
            ;; TODO: We potentially might have clashes, but its unlikely.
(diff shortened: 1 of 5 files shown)


vishesh:
Things seem to be working now. Tried some examples on Playground. Not sure if using `eval` is the best idea here, but effectively we were anyway doing that when taking playground input (as we were macro expanding anyway). Plus, if I understand correctly, it actually wont run module just create the module object.

I noticed that when switching b/w examples, sometimes the example wont load until I hit "Run" button again. That is probably an issue with playground client than the compiler itself.

stchang:
Nice, everything looks good and it's working great. Can we delete the `nothing.rkt` file now?

vishesh:
Yes. It is deleted in this PR. Merge whenever you are ready. (I dunno if I mentioned earlier, but you'd have to click approve from review to enable merge button).

stchang (approved):

Issue #84, opened 2017-08-29
Add support for `async` / `await`


amirouche:
In modern javascript, there is `async` and `await` new keywords that allows to write asynchronous code that looks like synchronous code. For instance,

```javascript
async function fetchPosts() {
    let response = await fetch('/api/posts');
    if (response.status === 200) {
         let posts = await response.json();
         return posts;
    }
}
```

Is it possible to have support for that? I don't how to port my framework to racketscript without this (see. #61).

The above should be possible to be written in racketscript using a syntax that looks like the following:

```scheme
(define-async (fetch-posts
 (let ((response (await (fetch "/api/posts"))))
       (if (eq? (response-status response) 200)
           (await (response-json response)))))
```

It should also be possible to define `async` lambda.

vishesh:
Sure I can do that. Is it ok if I do in weekend, or you are totally blocked?

amirouche:
It's ok I can wait.

jackfirth:
Could this be an ordinary Racket macro? As far as I know, it's a syntactic feature of JS only; there's no special runtime support for it.

vishesh:
@jackfirth Yes. That's what I originally intended. Otherwise just adding this to AST is probably a 10 min patch.

gregid:
+1 for this feature to be implemented

vishesh:
@amirouche Just wanted to do a quick update on this. I'm still working on this, but this will take some time before I get first prototype up. There is slight awkwardness with `yield` with recursive functions, and since Racket uses recursion for loops, the semantics needs to be right (preferably like Racket).

amirouche:
FWIW, I am not interested in `yield`. The only thing I need is the ability to declare a lambda `async` and call another lambda with `await`.

Hope that helps.

amirouche:
In BiwaScheme, I use a biwascheme primitive that rely on call/cc. In Javascript, I use `async`/`await`. FWIW [here](https://github.com/amirouche/socialite/blob/master/frontend/src/forward.js#L19) is the code that converts the async function into regular callback style function.

amirouche:
In BiwaScheme, I use a biwascheme primitive that rely on call/cc. In Javascript, I use `async`/`await`. FWIW [here](https://github.com/amirouche/socialite/blob/master/frontend/src/forward.js#L26) is the code that converts the async function into regular callback style function. The async function when called without `await` returns a promise.

gregid:
Is there any chance to implement this any time soon?

[...]

PR #224, opened 2021-06-18
htdp/universe: add mouse=?, make-color, bitmap/url

stchang:


stchang:
Do you remember why we didn't support `bitmap/url`? was it because of CORS blocking? I added it back because it seems to work (for some sites)

stchang:
(see the archery example I'm adding to the playground)

vishesh:
I think it was because we didn't do async/await and `bitmap-url` in Racket would be synchronous and JavaScript would be async. Speaking of which, we should probably think about async/await situation at some point.

vishesh (approved):

vishesh:
Just to keep a note, referencing this to https://github.com/vishesh/racketscript/issues/84 for async/await.

stchang:
Ah right. So it just might be slow? Is that causing this warning: `[Violation] 'load' handler took 157ms https://code.jquery.com/jquery-3.1.0.min.js`

vishesh:
Not sure if that message is related. Its asynchronous, the I think what we'd observe is image now showing up on screen, or some kind of undefined values flying around (when accessing image attributed --- unless they initialized to 0). I've not checked that particular error yet.

PR #228, opened 2021-06-22
enable ffi tests in fixture; closes #227

### LINKED ISSUE #227: fix and run ffi tests
I was looking into pr https://github.com/vishesh/racketscript/pull/226 and I realized the ffi tests are not running (and currently dont work)

stchang:


stchang:
for now, I just checked in expected output and compare against it

stchang:
When I click on codecov link to see the coverage I get:
```
Unauthorized
Github API rate limit error: Forbidden Check on Codecov's status or see our docs for common support.

Error 403
```

Do you get the same thing?

vishesh:
You need a Codecov account and connect it with Github (implied if you just use Github login to create account). The docs says:

> Codecov uses GitHub permissions to authenticate users. No setup necessary.
> Ask your team members to login to Codecov and they will have read/write access to the appropriate repositories.

stchang:
oh duh, haha

stchang:
If I'm reading it right, the `js-object?` and `js-array?` fns are not marked as covered even though I added tests for them.

I wonder if it doesn't like my macros?

vishesh:
I think makes sense, because `js-object?` is a runtime function. We are not running Racket version (just compiled JS) for this, so it never gets covered by Racket.

stchang:
So then we can't ever get coverage for those things?

vishesh:
I think, one way could be support sourcemaps[1]. That way we can get JavaScript coverage and map it to Racket lines. Its non-trivial work though. Sourcemap can also be useful in general to make programming experience with RacketScript much nicer (e.g. better stack-traces, potential browser integration etc?).

[1] https://github.com/vishesh/racketscript/issues/12

stchang:
This one I think need to fix, bc behavior of `js-object?` changed (https://github.com/vishesh/racketscript/pull/230/files)

tests/ffi/pr226.rkt.expected
```diff
@@ -0,0 +1,26 @@
+[object Object] is a js object?: #t
+[object Object] is a js array?: #f
+raw js array 1,2,3 is a js object?: #t
+raw js array 1,2,3 is a js array?: #t
+'#hash() is a js object?: #f
+'#hash() is a js array?: #f
+'#(1 2 3) is a js object?: #f
+'#(1 2 3) is a js array?: #f
+racket bytestring #"ABC" is a js object?: #t
```
stchang, line 9: @daynin @vishesh Is this proper behavior of `js-object?` Note that it still returns true for Racket bytestrings bc we directly use a typed array.

vishesh: Now that I remember, the original intention of `js-object?` was to just differentiate b/w object types and JS primitives, doesn't matter they came from. Its my bad for not asking the motivation for https://github.com/vishesh/racketscript/pull/230. 

Considering we named it `js-object`, perhaps we should treat it just as thin FFI function, which just returns exactly what JavaScript would think? I dunno what is the motivation in general to differentiate JS things from Racket things. I don't recall doing it anywhere.

stchang: I've never really used `js-object?` so I don't have a strong opinion. @daynin are you using `js-object?` to distinguish Racket objects from raw js?

stchang:
If we agree on the `js-object?` behavior, then this can be merged.

tests/ffi/pr226.rkt.expected
vishesh: I think we can just merge it. Can fix it later when we start looking at ffi again.

daynin: > I've never really used js-object? so I don't have a strong opinion. @daynin are you using js-object? to distinguish Racket objects from raw js?

(Sorry for not answering) Yes, sometimes we want to distinguish RacketScript primitives and JS ones

PR #230, opened 2021-06-24
fix: check if js object is not a racket primitive

daynin:
Without the fix it works like that:

```lisp
(js-object? ($/obj (test "test"))) // #t
(js-object? "test") // #t
```

### DIFF
--- a/racketscript-compiler/racketscript/interop.rkt
+++ b/racketscript-compiler/racketscript/interop.rkt
@@ -302,7 +302,11 @@
   (for/list ([(k v) (*in-js-object obj)]) (values k v)))
 
 (define (js-object? v)
-  (and ($/typeof v "object") (not (eq? v $/null))))
+  ($/binop &&
+    ($/binop &&
+      ($/typeof v "object")
+      ($/binop !== v $/null))
+    (not (($ ($ '$rjs_core) 'Primitive 'check) v))))
 
 (define (check-object v)
   (unless (js-object? v)


racketscript-compiler/racketscript/interop.rkt
```diff
@@ -302,7 +302,10 @@
   (for/list ([(k v) (*in-js-object obj)]) (values k v)))
 
 (define (js-object? v)
-  (and ($/typeof v "object") (not (eq? v $/null))))
+  (and
```
vishesh, line 305: Maybe we can try making this `($/binop '&& ...)` to avoid generating if/else statements and just use plain JS operators.

daynin: I'm not sure it'll be shorter. Looks like the only part which can be shortened like that is `(not (eq? v $/null))` -> `    ($/binop !== v $/null)`. Or I didn't get your idea correctly?

vishesh: E.g. (you can try at http://play.racketscript.org:8080)

```racket
#lang racketscript/base

(require racketscript/interop)

(define a (+ 1 2))
(define b (+ 2 3))

(displayln ($/binop && a b))
```

produces

```javascript
var a = 1 + 2;
var b = 2 + 3;
if (a !== false) {
  var if_res1 = b;
} else {
  var if_res1 = false;
}
M1.displayln(if_res1);
```

While 

```racket
#lang racketscript/base

(require racketscript/interop)

(define a (+ 1 2))
(define b (+ 2 3))

(displayln ($/binop && a b))
```

produces

```javascript
var a = 1 + 2;
var b = 2 + 3;
M2.displayln(a && b);
```

Difference b/w && and `and` is that `and` will aways produce a boolean, but `&&` may not (e.g.  `undefined && false`). But since in this case we know for sure both operands are boolean we can just use that. This is not common used function, but its runtime and is easy enough, so I guess why not.

daynin: Oh, I see. Ok, I'll try to rewrite it a little bit later. Thank you!

daynin: @vishesh done

The new code will produce a function like that:

```js
function(v1) {
  if (arguments.length !== 1) {
    throw $rjs_core.racketContractError("arity mismatch");
  } else {}
  return ((typeof(v1) === "object") && (v1 !== null)) && M1.not($rjs_core.Primitive.check(v1));
};
```

instead of:

```js
function(v2) {
  if (arguments.length !== 1) {
    throw $rjs_core.racketContractError("arity mismatch");
  } else {}
  if ((typeof(v2) === "object") !== false) {
    if (M1.not(M1.eq_p(v2, null)) !== false) {
      return M1.not($rjs_core.Primitive.check(v2));
    } else {
      return false;
    }
  } else {
    return false;
  }
}
```

We can also replace `not` with `$rjs_core.Primitive.check(v2) !== true` but i think it's not necessary.

vishesh: Thanks. Unrelated to this fix, I realized we should make `$/binop` take multiple args.

PR #231, opened 2021-06-25
feat: add `racket-list` for converting js arrays to racket lists

daynin:


racketscript-compiler/racketscript/interop.rkt
```diff
@@ -196,6 +197,9 @@
 (define (racket-string e)
   (($ ($ '$rjs_core) 'UString 'makeImmutable) e))
 
+(define (racket-list e)
```
vishesh, line 200: What do you think about naming this `js-array->list`? (like how generally Racket names conversions b/w types)

I see I've named previous function `racket-string`, and you're probably just following that, but we should probably change that as well.

daynin: Yes, I think it would be better

daynin: @vishesh done

vishesh (approved):

vishesh (approved):

PR #236, opened 2021-07-07
Tests for interop

gamburgm:
### What
Add some unit tests for the two JS interop modules.

### Why
Add some tests & code coverage while I'm spelunking through parts of the codebase.

### DIFF
--- a/racketscript-compiler/racketscript/boot/lang/private/interop.rkt
+++ b/racketscript-compiler/racketscript/boot/lang/private/interop.rkt
@@ -67,26 +67,30 @@
   (require rackunit
            racketscript/interop)
 
-  (define-simple-check (check-reader str first-id? result)
-    (equal? (syntax->datum
-             (read-racketscript #f (open-input-string
-                                    (string-append
-                                     (if first-id? "s." "s*.")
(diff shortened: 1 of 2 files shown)


racketscript-compiler/racketscript/boot/lang/private/interop.rkt
```diff
@@ -67,26 +67,30 @@
...
-
-  (check-reader "window" #t 'window)
-  (check-reader "window" #f '(#%js-ffi 'var 'window))
-
-  (check-reader "window.document" #t `(#%js-ffi 'ref window 'document))
-  (check-reader "window.document.write" #t
+  (define-simple-check (check-reader str expected)
+    (let ([actual (read-racketscript #f (open-input-string (substring str 2)))])
+      (equal?
+        (if actual
```
stchang, line 73: Is this `if` actually needed?

gamburgm: Yes, it's just to check when `actual` may be `#f`, and when it is `syntax->datum` blows up with a contract violation.

stchang (approved):

PR #239, opened 2021-07-08
Update readme.md

kikudjira:
I added a logo (maybe it will suit for you) and corrected minor places in the text.


vishesh:
Nice, I like it! Specially the red/blue gradient border. Thanks for making it!! 

However, I'm not sure about using Racket logo itself in it though for two reasons (1) It is a logo owned by Racket project, so copyright issues (2) Some people may not like us so aggressively trying to say we are Racket and it is totally justified because we are not (we try to be though).

Also, how about instead of `RS` we write `(rs)` i.e. with parenthesis? If the parenthesis one doesn't look good, I think `RS` looks great too.

@stchang What do you think?

stchang:
I like it too, both the proposal to add a logo and the design itself. Thank you for this contribution!

I also had some concern about licensing/copyright issues, but this is an area I don't know much about. ([I found this](https://commons.wikimedia.org/wiki/File:Racket-logo.svg))

I'm mostly ok with any design, but I think @vishesh's suggestions to make it more derivative is probably a good idea, just to avoid potential problems down the road.

vishesh (approved):

vishesh:
Awesome! Thanks for making the changes. Looks great!

@stchang: What do you think? If you are ok with the design, we can merge this.

kikudjira:
@vishesh @stchang thank you for your comments! 

What do you think of these [logo variations](https://s.csssr.ru/U1A4UD346/racketscript_logos.png)? I like the last two. The arrow, according to my plan, was to reflect the possibility Racket → JS :-)

https://github.com/kikudjira/racketscript

vishesh:
My favorites are 5th and 6th one (first two of last three).

kikudjira:
@vishesh I'm glad you liked it! 

You may need these files (wb — with white background, you can use for avatars in this organization):
https://s.csssr.ru/U1A4UD346/rs_logo_1.svg
https://s.csssr.ru/U1A4UD346/rs_logo_1_wb.svg
https://s.csssr.ru/U1A4UD346/rs_logo_2.svg
https://s.csssr.ru/U1A4UD346/rs_logo_2_wb.svg

stchang:
Wow @kikudjira these are really great! Thanks again for creating these.

I also like 5th and 6th (same as @vishesh) with the gradients.

If others like the arrow, we can go with the gradient arrow (that's the one currently used in this pr already I think?)

Yes, let's merge this.

stchang (approved):

vishesh:
We should add the other 4 files somewhere too, or else we'll lose them. Should we add an `assets/logo` directory in this repo or create another repo for any assets we may have related to the project?

stchang:
> Should we add an `assets/logo` directory in this repo or create another repo for any assets we may have related to the project?

Either sounds good

kikudjira:
@vishesh I'm sorry, but It seems to me that for the organization page it is better to use this file https://s.csssr.ru/U1A4UD346/rs_logo_2_wb.svg. Otherwise, in the dark theme, the github does not look very nice: https://s.csssr.ru/U1A4UD346/Screenshot_GH.png

vishesh:
Cool. I just updated the org picture.

PR #247, opened 2021-07-26
WIP: Replace Symbol class with native Symbol

### LINKED ISSUE #246: compile Racket symbols to js Symbol values?
js has Symbol values now: https://developer.mozilla.org/en-US/docs/Glossary/Symbol

Would it be appropriate to compile Racket symbols to these js values?

(@samth Do you know if these would be compatible with Racket symbols, eg interning?)

DarrenN:
Fixes #246

⚠️ **Very exploratory**

Reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol)

Swaps in native Symbol to provide interned (`Symbol.for()`) and uninterned (`Symbol()`) symbols. `/runtime/core/symbol.js` is now mostly a collection of utils to create and compare values.

JS Symbols print like "Symbol(foo)" when `.toString` is called on them. To correctly pass the integration tests and mimic Racket's printing behavior we need to pull the value from interned Symbols and display that instead. This adds some utils to make it easier to check if a value is a `Symbol` and correctly return the type of string we want to use in display.

## Notes

- Because native `Symbol`s don't implement many of the methods from `Primitive` and friends we have to do a lot of special checks in lots of different places (ex: `struct.js`). Is this "ok" or should we consider wrapping the interned `Symbols` in an interface anyway to handle printing issues?
- I didn't change the way the transforms treat `symbol?` as my first swing at it caused a lot of issues. We can always go back and re-consider.

### DIFF
--- a/racketscript-compiler/racketscript/compiler/runtime/core/errors.js
+++ b/racketscript-compiler/racketscript/compiler/runtime/core/errors.js
@@ -1,3 +1,4 @@
+import { safeToString } from './primitive.js';
 import { MiniNativeOutputStringPort } from './mini_native_output_string_port.js';
 import { printNativeString } from './print_native_string.js';
 // this (errors.js) depends on UString, so vice versa cannot be true
@@ -10,16 +11,21 @@ function printError(out, msg, args) {
         if (typeof arg === 'string') {
             out.consume(arg);
         } else {
-            printNativeString(out, arg, /* printAsExpression */ true, /* quoteDepth */ 0);
+            printNativeString(
+                out,
+                arg,
+                /* printAsExpression */ true,
+                /* quoteDepth */ 0
+            );
         }
     }
 }
 
 function makeError(name) {
     /**
-     * The "(error msg v ...)" form.
-     * Besides Racket values, also allows native strings.
-     */
+   * The "(error msg v ...)" form.
+   * Besides Racket values, also allows native strings.
+   */
     const e = function (msg, ...args) {
         this.name = name;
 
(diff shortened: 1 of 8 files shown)


samth:
Am I right that all the places that use `.toString()` that change here will eventually need to change to some more correct string conversion?

Also, can Symbol be subclassed?

DarrenN:
@samth - We can't subclass native objects in JS, but we can wrap them (either with a `class` or old school prototypes) to extend with our own methods, such as a custom version of `toString`. Which I think would be a better approach here tbh. I'll work up some examples.

DarrenN:
Actually I take that back we can `extend` Symbol I think https://javascript.info/extend-natives

stchang:
> Actually I take that back we can `extend` Symbol I think https://javascript.info/extend-natives

Cool, I didn't know this either. I just assumed we would have to extend `print-native-string` like you did

DarrenN:
@stchang it's still a little wonky because `Symbol` uses `static` methods: https://jsbin.com/nizipuc/1/edit?js,console

```

class PrimitiveSymbol extends Symbol {
  constructor(name) {
    super (name);
    this.sym = Symbol.for(name);
  }

  toString() {
    return Symbol.keyFor(this.sym);
  }

}

let sym = new PrimitiveSymbol("fred");
console.log(sym.toString())

let sym2 = new PrimitiveSymbol();
console.log(sym2.toString())
```

DarrenN:
Superseded by #250

PR #250, opened 2021-08-02
Replace Symbol with PrimitiveSymbol using native JS Symbols under the hood

### LINKED ISSUE #246: compile Racket symbols to js Symbol values?

DarrenN:
Fixes #246 
Related to #247 

Replaces `core/symbol.js` with `core/primitive_symbol.js` which inherits from `PrintablePrimitive` and `Primitive`. The new class prevents overriding the native `Symbol` while using `Symbol.for` under the hood for interned symbols and `Symbol` for the uninterned. 

While we can _kind of_ subclass `Symbol` we would still have issues with the different printing methods that `PrintablePrimitive` provide. This would require lots of cascading changes to the printing modules to specifically check for a Symbol-like object.


DarrenN:
Hmmm.. looks like tests on Racket 6.12 BC are on the struggle bus. Any tips? Is there a docker container with this version I can test on locally?

vishesh:
Don't have a docker image in this repo, but you can modify this Dockerimage from playground: https://github.com/racketscript/racketscript-playground/blob/master/Dockerfile

@stchang worked quite a bit on supporting various version, so he may have more tips.

stchang:
I just have a test utility macro wrapper, `run-if-version`, where it controls whether a test runs based on the version.
Here's an example: https://github.com/racketscript/racketscript/blob/master/tests/racket-core/hash.rkt#L372

If a test is failing only in 6.12, if you can find which test it is, then I would just add a `(run-if-version "7.0" ....)` around it

vishesh (approved):

stchang:
Hurray! Thanks for the contribution!

PR #254, opened 2021-08-12
Allow for `#f` default value

### LINKED ISSUE #252: make-vector default value doesnt allow false
`(make-vector 3 #f)` should evaluate to a vector of false values but instead it mistakenly uses the default `0` value

http://play.racketscript.org/#gist/c5fc2837855bbebacea191f255b5a251

gamburgm:
Fixes #252.

Previously, if the default value provided to `make-vector` was falsey, the vector would be populated with `0`. Now this will only happen if the default value is `undefined`.

If we wanted to enable a developer to populate a vector with `undefined`, we'd have to be able to detect if a value was passed in to the function or not. We could do so by accessing `arguments` directly if we feel like this is something a user should be able to do. 

As an addendum, even though `kernel.rkt` is enormous, might still be good to create a test submodule and throw things in there as people modify it.

### DIFF
--- a/racketscript-compiler/racketscript/compiler/runtime/kernel.rkt
+++ b/racketscript-compiler/racketscript/compiler/runtime/kernel.rkt
@@ -394,6 +394,7 @@
 (define-checked+provide (vector->immutable-vector [vec vector?])
   (#js.Core.Vector.copy vec #f))
 
+
 ;; --------------------------------------------------------------------------
 ;; Hashes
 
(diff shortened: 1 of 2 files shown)


sorawee:
In Racket, an approach I usually use is to generate a new unique object (via `(gensym)`, `(lambda () #f)`, or even `(cons #f #f)`), use it as the default value, and use `eq?` to test that users do provide arguments.

Does that work here?

gamburgm:
> In Racket, an approach I usually use is to generate a new unique object (via `(gensym)`, `(lambda () #f)`, or even `(cons #f #f)`), use it as the default value, and use `eq?` to test that users do provide arguments.
> 
> Does that work here?

I think some infrastructure would need to change to enable that, you can't provide default values for parameters currently for built-in functions. Not 100% certain though.

gamburgm:
@stchang a `test` submodule is a little tough because `kernel.rkt` calls the FFI. I'll just add tests to `tests/` for now.

stchang (approved):

stchang:
Thanks!

PR #266, opened 2021-11-12
Compile Linklets to Javascript

gamburgm:
@stchang 

We would like to eliminate the server requirement for editing Racketscript online. To do that, we need to be able to run the Racketscript compiler itself (and more importantly, the Racket expander) in the browser.

The first step in achieving this is compiling linklets to javascript. The expander can compile itself to one, so this is the first milestone in the bootstrapping process.

Warning: the code is a mess. The code was written under the (reasonable) assumption that there wouldn't be another module-like entity to compile, but I've skipped refactoring any of that at the moment.

### DIFF
--- a/stuff.rkt
+++ b/stuff.rkt
@@ -1,0 +1,6 @@
+#lang racket/base
+
+(+ 2 3)
+
+38
+
--- a/racketscript-compiler/racketscript/compiler/absyn.rkt
+++ b/racketscript-compiler/racketscript/compiler/absyn.rkt
@@ -5,6 +5,8 @@
 
 (provide (all-defined-out))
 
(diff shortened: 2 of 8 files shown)


stuff.rkt
```diff
@@ -0,0 +1,6 @@
+#lang racket/base
```
pmatos, line 1: This file probably shouldn't be committed.

pmatos: but apparently there are others so this might be still early WIP, in which case disregard. :)

gamburgm: Yeah this is still in early stages, some of these garbage files (especially `expander.rktl`) are just convenient when I'm switching computers or someone else wants to debug something -- hopefully this is in a better state soon :^)

PR #282, opened 2021-12-16
[WIP] complex numbers

stchang:
see https://github.com/racketscript/racketscript/issues/271


racketscript-compiler/racketscript/compiler/transform.rkt
```diff
@@ -665,6 +665,11 @@
          (void? d)
          (real? d))
      (ILValue d)]
+    [(and (number? d) (zero? d)) (ILValue d)] ; some zero values are not real, eg 0.0+0.0i
+    [(and (number? d) (not (real? d))) ;; FIXME with real complex num implementation
+     (ILApp (name-in-module 'core 'Pair.make)
```
vishesh, line 670: Should Complex type be a class, so that we can differentiate b/w complex and pair of reals.

PR #294, opened 2022-01-11
`prop:sealed` support

gamburgm:
Adds `prop:sealed` as a struct type property, as well as implementing its semantics (no other struct type can have a sealed struct type as super). 

This property is used in the expander linklet.


stchang (approved):

samth:
Note that the implementation here isn't really right; this struct is not sealed:

```
(define-values (p:s p? p-acc) (make-struct-type-property 'prop:sealed))
(struct x () #:property p:s #t)
(struct y x ())
```

gamburgm:
Got it, I'll put up a PR tomorrow to correct that.

stchang:
Can you also add that example as a test?

Issue #313, opened 2022-10-04
Module dependencies are not always recompiled when updated


kclapper:
If a change is made to a module's dependency, but the module itself is left alone, the dependency is not recompiled.

For example, if you have the file `has-dependency.rkt`:
    
    #lang racket 
    
    (require "dependency.rkt")

    (println (format "~a" (add 5)))

and another file `dependency.rkt`:

    #lang racket 
    
    (provide add)

    (define (add n)
      (+ n 2))

If you compile `has-dependency.rkt`, then change the `2` in `dependency.rkt` to a `1` and recompile it again, the resulting Javascript code still prints out "7" instead of "6".

kclapper:
This can be fixed by checking for stale modules in the cached timestamps ahead of the compilation loop in `racket->js`. Then, the stale modules can be added to the queue to be recompiled. This solution seems to work for me, I'll try and write tests and submit a pull request soon.

stchang:
fixed with #314

PR #314, opened 2022-10-18
Fix stale dependency compilation issue.

kclapper:
Fixes the issue where a stale dependency is not recompiled if the dependent file hasn't changed. It does this by checking all cached timestamps for stale files before the main compilation loop, instead of doing it in flight.

This resolves issue #313.

I'm having some trouble writing a test case for this code. There's a difference between the main compiler and the test fixture that causes this issue to crop up in tests despite the fix.  When using `racks`, this fix produces the desired behavior. I'll push another commit and update this pull request once I have a working test case.

### DIFF
--- a/racketscript-compiler/racketscript/compiler/main.rkt
+++ b/racketscript-compiler/racketscript/compiler/main.rkt
@@ -250,12 +250,12 @@
     (when (not (skip-module-compile? timestamps mod))
       (put-to-pending! mod)))
 
-  (let loop ()
+  (let loop ([compiled-modules empty])
     (define next (and (non-empty-queue? pending) (dequeue! pending)))
     (cond
       [(and next (skip-module-compile? timestamps next))
        (log-rjs-info (~a "Skipping " next))
-       (loop)]
+       (loop compiled-modules)]
       [next
        (current-source-file next)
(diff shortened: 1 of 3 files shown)


kclapper:
What I settled on for a test case is to have a simple racket file that `require`s another simple file. The test simply changes the required file's last modification timestamp so it shows up as a stale dependency. Then the test checks to make sure the compiler recompiles it. 

To make this work, I made a slight change to the `racket->js` function so that it returns a list of the modules it compiled. Typically this function is used solely for it's side effects, so this change doesn't break any existing tests. The function can still be used normally while ignoring the list output. Having some form of output from this function may come in handy later too, perhaps in other tests.

Please let me know if anyone has suggestions, I'd be happy to go back and make improvements.

tests/fixture.rkt
```diff
@@ -317,6 +317,8 @@
   (run tc-search-pattern))
 
 (module+ test
+  (require rackunit)
```
stchang, line 320: is this needed? or is it leftover from some experimentation?

kclapper: Oops, yes that was from some experimentation. I just fixed it and amended the commit. Thank you.

stchang (approved):

PR #315, opened 2022-11-14
Fix unintended printing of module paths to stdout

kclapper:
After #314 was merged, I noticed when the compiler finished it would print a list of compiled modules to standard out, which is unintended behavior. This PR fixes the issue.

In this PR I also took the opportunity to refactor the `(racket->js)` method.

### DIFF
--- a/racketscript-compiler/racketscript/compiler/main.rkt
+++ b/racketscript-compiler/racketscript/compiler/main.rkt
@@ -221,70 +221,64 @@
        (= (get-module-timestamp ts mod)
           (file-or-directory-modify-seconds (actual-module-path mod)))))
 
-;; -> Void
+;; -> (Setof Path)
 ;; For given global parameters starts build process starting
-;; with entry point module and all its dependencies
+;; with entry point module and all its dependencies. Returns
+;; a set module paths that were compiled (typically ignored).
 (define (racket->js)
   (define added (mutable-set))
   (define pending (make-queue))
-
-  ;; build directories to output build folder.
-  (define default-module-name (string-slice (~a (last-path-element
-                                                 (main-source-file)))
-                                            0 -4))
-  (prepare-build-directory default-module-name)
-
   (define (put-to-pending! mod)
     (unless (set-member? added mod)
       (set-add! added mod)
       (enqueue! pending mod)))
 
-  (define timestamps (load-cached-module-timestamps))
+  (define default-module-name (string-slice (~a (last-path-element
+                                                 (main-source-file)))
+                                            0 -4))
+  (prepare-build-directory default-module-name)
 
   (put-to-pending! (path->complete-path (main-source-file)))
   (for ([pm primitive-modules])
     (put-to-pending! pm))
 
-  ;; Add stale modules to the compile queue.
+  (define timestamps (load-cached-module-timestamps))
   (for ([(mod timestamp) timestamps])
     (when (not (skip-module-compile? timestamps mod))
       (put-to-pending! mod)))
 
-  (let loop ([compiled-modules empty])
-    (define next (and (non-empty-queue? pending) (dequeue! pending)))
-    (cond
-      [(and next (skip-module-compile? timestamps next))
-       (log-rjs-info (~a "Skipping " next))
-       (loop compiled-modules)]
-      [next
-       (current-source-file next)
-       (make-directory* (path-only (module-output-file next)))
-       (save-module-timestamp! timestamps next)
+  (define compiled-modules
+    (for/set ([next (in-queue pending)]
+              #:unless (and (skip-module-compile? timestamps next)
+                            (log-rjs-info (~a "Skipping " next))))
 
-       (define expanded (quick-expand next))
-       (define ast (convert expanded (override-module-path next)))
+      (current-source-file next)
+      (make-directory* (path-only (module-output-file next)))
+      (save-module-timestamp! timestamps next)
 
-       (assemble-module (insert-arity-checks
-                         (absyn-module->il* ast))
-                        #f)
+      (define expanded (quick-expand next))
+      (define ast (convert expanded (override-module-path next)))
+      (assemble-module (insert-arity-checks
+                        (absyn-module->il* ast))
+                       #f)
 
-       ;; Run JS beautifier
-       (when (js-output-beautify?)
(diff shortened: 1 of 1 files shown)


racketscript-compiler/racketscript/compiler/main.rkt
```diff
@@ -221,54 +221,48 @@
...
 
-  (define (put-to-pending! mod)
-    (unless (set-member? added mod)
-      (set-add! added mod)
-      (enqueue! pending mod)))
-
   (define timestamps (load-cached-module-timestamps))
+  (for ([(mod timestamp) timestamps])
+    (when (not (skip-module-compile? timestamps mod))
+      (put-to-pending! mod)))
```
stchang, line 244: Moving this here changes the order that the dependencies are visited? Is that the intention? Or this is just a code readability refactoring?

racketscript-compiler/racketscript/compiler/main.rkt
```diff
@@ -221,54 +221,48 @@
...
 
-  ;; Add stale modules to the compile queue.
-  (for ([(mod timestamp) timestamps])
-    (when (not (skip-module-compile? timestamps mod))
-      (put-to-pending! mod)))
-
-  (let loop ([compiled-modules empty])
-    (define next (and (non-empty-queue? pending) (dequeue! pending)))
+  (define compiled-modules (mutable-set))
+  (for ([next (in-queue pending)])
```
stchang, line 251: good refactor

stchang: Would `for/set` work here? Then you could eliminate the explicit `set-add!`s (does `compiled-modules` need to be mutable?)

racketscript-compiler/racketscript/compiler/main.rkt
kclapper: Yeah it was just supposed to be readability refactoring. I didn't notice that putting it before the loop that adds the primitive modules changes the initial order of the compilation queue. I don't think it changes the behavior of the output JavaScript but it probably makes more sense to move this back to where it was. I'll just also move the line where we bring in the cached timestamps.

racketscript-compiler/racketscript/compiler/main.rkt
kclapper: I didn't know about `for/set`, it would totally work. The reason I was using a mutable set was just to maintain the same convention as the `pending` queue and `added` set.

racketscript-compiler/racketscript/compiler/main.rkt
```diff
@@ -221,70 +221,65 @@
...
-       (dump-module-timestamps! timestamps)
-       (unless (equal? (js-target) "plain")
-         (log-rjs-info "Running NPM [Install/Build].")
-         (npm-install-build))
-       (log-rjs-info "Finished.")
-       compiled-modules])))
+  (define compiled-modules
+    (for/set ([next (in-queue pending)])
+      (cond
+        [(skip-module-compile? timestamps next)
```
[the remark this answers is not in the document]
kclapper: That would be a good change. If we use `#:unless`, we lose the ability to log that we skipped a module. We could move this logging statement to `skip-module-compile?` but it seems like it's better to leave it in the body of `racket->js` somehow. Is there a clean way to have the logging be a part of the `#:unless` clause or should it just be moved to `skip-module-compile?`?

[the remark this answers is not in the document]
kclapper: Wait, never mind, I think I found a good solution. I'll push a commit in a minute.

racketscript-compiler/racketscript/compiler/main.rkt
```diff
@@ -221,70 +221,64 @@
...
-       (dump-module-timestamps! timestamps)
-       (unless (equal? (js-target) "plain")
-         (log-rjs-info "Running NPM [Install/Build].")
-         (npm-install-build))
-       (log-rjs-info "Finished.")
-       compiled-modules])))
+  (define compiled-modules
+    (for/set ([next (in-queue pending)]
+              #:unless (and (skip-module-compile? timestamps next)
+                            (log-rjs-info (~a "Skipping " next))))
```
stchang, line 253: good

racketscript-compiler/racketscript/compiler/main.rkt
```diff
@@ -221,70 +221,64 @@
...
+                        (absyn-module->il* ast))
+                       #f)
+
+      (when (js-output-beautify?)
+        (system (format "js-beautify -r ~a" (module-output-file next))))
+
+      (for ([mod (in-set (Module-imports ast))])
+        (match mod
+          [(? symbol? _) (void)]
+          [_ #:when (collects-module? mod) (void) (put-to-pending! mod)]
```
stchang, line 271: Is this `(void)` needed? (I know you didnt add it)

racketscript-compiler/racketscript/compiler/main.rkt
```diff
@@ -221,70 +221,64 @@
...
+          [_ (put-to-pending! mod)]))
```
stchang, line 272: So the only iffy thing now is that we are mutating the `queue` while iterating. I think it will work, but only because it's a queue so things get added to the end

stchang: I had to look at the [implementation of `queue`](https://github.com/racket/racket/blob/master/racket/collects/data/queue.rkt) to be sure

racketscript-compiler/racketscript/compiler/main.rkt
kclapper: If you take out the first `(void)` you get an error from `match` but I was able to take out the second `(void)` with no problems. Based on the syntax of `match` it seems to be serving no purpose so I'll leave it out.

racketscript-compiler/racketscript/compiler/main.rkt
kclapper: Oh good, thank you for doing that. I went looking for its implementation but couldn't find it before having to go to class. I had a feeling it would be fine because all of the integration tests pass and some of them have `require`d modules, but definitely better to see for certain. I agree it's otherwise somewhat iffy.

stchang (approved):

